PR #15, opened 2015-05-27
Feature/polygon area

martiwer:


pfankhauser:
Looks good thanks. What do you use the GetGridMapInfo.srv service for?

martiwer:
I use the GridMapInfo to set the bounds of the bounds of the planner to the bounds of the traversability map. The service updates the traversability map and gets the infos of this map.

PR #17, opened 2015-06-11
grid map -> point cloud conversion and publication is only executed w…

remod:
grid map -> point cloud conversion and publication is only executed when a subscriber is around


pfankhauser:
Not necessary, visualization is not active if not subscribed, check `VisualizationBase::isActive()`.

PR #21, opened 2015-07-22
Added spiral iterator.

martiwer:


grid_map_core/src/iterators/SpiralIterator.cpp
```diff
@@ -0,0 +1,123 @@
...
+int SpiralIterator::dist(const double dx, const double dy)
+{
+  return (int) std::round(sqrt(dx*dx + dy*dy));
+}
+
+bool SpiralIterator::isInside(const Index index)
+{
+  Eigen::Vector2d position;
+  getPositionFromIndex(position, index, mapLength_, mapPosition_, resolution_, bufferSize_);
+  double squareNorm = (position - center_).array().square().sum();
```
pfankhauser: Use `squaredNorm()` Eigen method instead.

PR #31, opened 2016-01-31
Install the image publisher script into the runtime directory

stonier:
Demo image launcher failing without this.


pfankhauser:
Thanks a lot!

PR #35, opened 2016-02-11
Fix conversion to OccupancyGrid msg

czalidis:
* Use the unwrapped index when converting to one dimensional matrix indexing.
* Add a test case that demonstrates the above situation. The test should fail
  if used without the unwrapped index.
* The bug only appears if the map has been moved before the conversion. (so there is an offset to the circular buffer)

The test case only tests for this specific bug. Let me know if I should add more tests or if I should discard it altogether.


pfankhauser:
Thanks a lot, I merged this work to the branch `feature/occupancy_to_grid_map`. Could you please test if the changes in that branch work for you? If you have any more unit tests, feel free to do another pull request.

czalidis:
The changes seem to work, the only comment that I have is whether the OccupancyGrid unknown value (`-1.0`) should be translated to `NAN` or not. I left an inline note about that [here](https://github.com/ethz-asl/grid_map/commit/e359355680322595921924b2514e12708bbadd37).

Also, now that you have implemented the `fromOccupancyGrid()` function, it sounds reasonable to add a unit test that does a round-trip conversion for OccupancyGrid to GridMap and back.

Thanks for the quick fix!

PR #36, opened 2016-02-12
Feature/occupancy to grid map

pfankhauser:


pfankhauser:
@czalidis Thank you for pointing out the inconsistency. I fixed the `NAN` to `-1` conversion and added a round-trip unit test. Could you please recheck the latest code?

czalidis:
New changes work as expected for my use case. Thanks!

pfankhauser:
Thanks @czalidis!

PR #39, opened 2016-02-18
Const correctness for getSubmap

stonier:
So we can use it when grid maps are passed in as const references to functions.


pfankhauser:
Thanks!

PR #40, opened 2016-02-20
Templatise MultiArray <-> Eigen Matrix converters

stonier:
There is already one templatised method - this just applies the same
methodology to the rest of the MultiArray API so we can reuse it with
the cost_map implementation (grid_maps with unsigned chars).

This is the first step I'd like to make in reusing as much as possible between the float and unsigned char implementations.


pfankhauser:
Great, thanks a lot for your work. I extended it a bit and replace this PR with #43.

PR #41, opened 2016-02-20
getIndexFromPosition() should not return index larger than bufferSize

czalidis:
In my first commit I added a test case that tests the return value of `getIndexFromPosition()` when a position right on the edge is given. Turns out that the function returns `true` and the returned index is invalid (index is equal to bufferSize). I noticed that `checkIfPositionWithinMap()` accepts positions that are exactly on the edge of the map but when the index is generated there is no check to validate that the index is still in range.

My solution was to modify `checkIfPositionWithinMap()` and not accept positions on the edge of the map. It seems that this change doesn't affect anything else.

Another solution (if it is desired to accept edge positions) would be to make sure that the returned index will be in-bounds, in similar fashion as the `limitPositionToRange()` does. In this case, the test that I added in my first commit should pass.

Since I am not sure about the desired behavior of `checkIfPositionWithinMap()` regarding edge cases, let me know if something else is indeed affected by this change.


pfankhauser:
I don't think there's an expected behavior for these edge cases so I would go ahead with your proposed solution. Thanks!

PR #42, opened 2016-02-21
Method for setting the position

stonier:
Right now the only place to set the position seems to be via `setGeometry`, which takes care of cleaning out all the layers etc.

We have a use case where the map width and height are fixed, and an update can happen just by writing into existing storage which is much faster than deleting it all and recreating. However, it does need a way to be able to update the position which was not hitherto possible.

Am I missing something in the usage of the library, or have you not used it in this way before?

### DIFF
--- a/grid_map_core/include/grid_map_core/GridMap.hpp
+++ b/grid_map_core/include/grid_map_core/GridMap.hpp
@@ -71,6 +71,12 @@ class GridMap
   void setGeometry(const SubmapGeometry& geometry);
 
   /*!
+   * Set the position of the grid map.
+   * @param position the 2d position of the grid map in the grid map frame [m].
+   */
+  void setPosition(const Position& position) { position_ = position; }
+
+  /*!
    * Add a new empty data layer.
    * @param layer the name of the layer.
    * @value value the value to initialize the cells with.


pfankhauser:
Have you looked at the `move(const Position& position)` method? This changes the position of the map without resetting any data. However, in our typical use cases the map data is associated to the an inertial reference system, so we use the `move(...)` method which essentially moves the map boundaries while leaving the data associated with the same location in the grid map frame. In case there's a use case where you really want to change the position of the map without altering the map data, your addition would make change.

stonier:
From what I can see, in a large move it is still doing alot of marking (`data_.at(layer).setConstant(NAN);`) which I would then have to rewrite with new data, i.e. two operations where one would be minimal. 

I have to admit, I haven't got a good handle on exactly what you are doing with the move operation, nor have I been able to get it to produce what I would expect from it. I suspect accessing the underlying matrix directly is misleading once the boundaries of the grid map have moved...I'll start doing some simple experiments and get a better idea of what is going on.

pfankhauser:
I have added a new method `setPosition` as suggested. I have also added a new demo to illustrate the difference between the `move` and the new `setPosition` method. You can run the demo with:

    roslaunch grid_map_demos move_demo.launch

pfankhauser:
[This](https://github.com/ethz-asl/grid_map/blob/master/README.md#changing-the-position-of-the-map) should further clarify the different methods to change the position of the map.

stonier:
Thanks! And nice visualisation...one of those things that transmits an idea immediately :neckbeard:

With this issue, I'm getting the hang of using the library and iterators in the intended fashion now.  It's proving to be very useful and saving alot of code. Nice job.

PR #47, opened 2016-03-03
Visualise flat point clouds

stonier:
For double maps, converting to an occupancy grid map is a right nuisance inasmuch as you have to take care to convert values within 'proper' cost value bounds.

This PR provides a way to construct a 'flat' point cloud, that will the subsequently show the values in the layer via the colour derived from the grid map values stored in a channel.

Example yaml for the visualization node:

```
grid_map_topic: /planner/potential_grid_map
grid_map_visualizations:
  - name: potentials
    type: point_cloud
    params:
      layer: normalised
      flat: true
```

Some points to note:

* This changes the `toPointCloud` api, could drop the boolean down and give it a default argument, but that would alter the style of the rest of the functions for which `const` args always precede.

* Should the visualisation dump *all* other layers into the channels by default? Via the yaml, the impression is that I'm instructing it only to dump one layer.

### DIFF
--- a/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp
+++ b/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp
@@ -77,9 +77,12 @@ class GridMapRosConverter
    * additional fields.
    * @param[in] gridMap the grid map object.
    * @param[in] pointLayer the type that is transformed to points.
+   * @param[in] flat flatCloud, set z = 0 and values in the channels (useful for 2d maps in rviz).
    * @param[out] pointCloud the message to be populated.
    */
-  static void toPointCloud(const grid_map::GridMap& gridMap, const std::string& pointLayer,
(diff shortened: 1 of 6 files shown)


stonier:
flat_point_cloud

pfankhauser:
Thanks for your input. However, I'm not sure if it's a good idea to add a specific use case to a general implementation. The same thing can be solved with adding a new layer with a constant value which then can be used as the position layer:

    map["normalised"].setZero();

or

    map_.add("normalised", 0.0);

This way you could also specify the height at which you want to show the flat point cloud. We have used this approach to show the projected images of a streaming camera in Rviz.

image

stonier:
Yes, I'd actually done like this within the layers prior to looking at this.

I would, however, like to leave rviz related tricks outside of the main program (I really like the way the visualisation node does alot of this for you). I could put a node between the main program and the visualization node, but that is another node...

pfankhauser:
I see your point. Could you please check if #48 works for you?

PR #48, opened 2016-03-04
Flat point clouds

pfankhauser:
#47

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ The published topics are configured with the [YAML parameter file](grid_map_demo
 
 * **`point_cloud`** ([sensor_msgs/PointCloud2])
 
-    Shows the grid map as a point cloud. Select which layer to transform as points with the `layer` parameter. Set the optional parameter `flat` to true to show the point cloud with height *z* = 0.
+    Shows the grid map as a point cloud. Select which layer to transform as points with the `layer` parameter.
 
         name: elevation
         type: point_cloud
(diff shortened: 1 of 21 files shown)


stonier:
Conceptually has one difference to #47 - it publishes all cells in the grid map into the cloud.

I think it is preferable to drop any points for which `isValid()` is not true.

stonier:
Visualisation behaviour here changes (compared with #47) for *invalid* points. Previously,

```
x, y, z = nan, nan, \inf
```

This caused them to be skipped in a point cloud visualisation, which lets an autocomputed rainbow visualisation via intensities work as intended. Essentially this is due to the following lines in `GridMapRosConverter::toPointCloud`

```
    Position3 position;
    position.setConstant(NAN);
    gridMap.getPosition3(pointLayer, *mapIterator, position);
```

By introducing a completely zero layer, this means all points are now valid and they get inserted into the point cloud as

```
x, y, z = position.x(), position.y(), \inf
```

which causes the autocomputed rainbow visualisation to include the inf's as some very high (perhaps float max) value. 

Perhaps the right workaround is to iterate over the original layer and switch values in the 'flat' layer to `/inf` where appropriate?

pfankhauser:
Thanks for the info, I see your point. I've updated the code to accommodate for this behavior. I have also added a new `flat_point_cloud` visualizer to separate things clearer. This is documented [here](https://github.com/ethz-asl/grid_map/tree/flat_point_clouds#published-topics).

So in your case, you should add the layer which for which you wish to skip the `NAN`-cells in the conversion as basic layer with

    gridMap.setBasicLayers({"your_layer"});

Could you please check if these changes now work for you?

stonier:
Big +! from me! :neckbeard:

Especially nice being able to set the height parameter - I can now drop the point cloud to just below where the other renderings happen.

pfankhauser:
Cool, thanks! I'll merge then.

PR #58, opened 2016-03-28
Fix mixed up index for cv image access (#56)

skohlbr:
See #56. Supersedes #57, which (sort of) fixed things at the wrong place.


pfankhauser:
Superseded by #61.

PR #59, opened 2016-04-07
Makes the isInside checks const

stonier:
So it can be used in const functions.


bit-pirate:
Discussion about how to best add this feature is over here: https://github.com/ethz-asl/grid_map/issues/60

pfankhauser:
Thanks!

PR #65, opened 2016-08-18
Feature/rviz plugin

pfankhauser:
Introduces new Grid Map RViz plugin to visualize grid maps as height reliefs in RViz.


pfankhauser:
Thanks @kruesip!

PR #73, opened 2016-10-21
fixed eigen plugin compiler error

djud:


Ryanf55:
Can you provide context for this change, how to reproduce the original issue, and why the changes were needed? We still have open issues regarding the build issues for the plugins and I'm not sure what's going to break or how to test any changes to the definitions added in this PR.

Issue #76, opened 2016-11-11
Conversion from CostMap2D to GridMap


skohlbr:
Standard ROS navigation uses [CostMap2D](https://github.com/ros-planning/navigation/blob/0fec250ead4a8b2c5ed2e6d1e3efb28e7885c12e/costmap_2d/src/costmap_2d.cpp) as a cost map representation and when implementing a [nav_core::BaseGlobalPlanner](https://github.com/ros-planning/navigation/blob/0fec250ead4a8b2c5ed2e6d1e3efb28e7885c12e/nav_core/include/nav_core/base_global_planner.h#L48) plugin that uses grid_map as a back-end, it would be desirable to have a converter that allows initalizing the internal grid_map from a CostMap2D.
Are there any plans for implementing this (or even some implementation out there that I missed)?

stonier:
We have a function in the `cost_map` library which does this...[here](https://github.com/stonier/cost_map/blob/devel/cost_map/src/lib/converter.cpp#L347). There's an active [PR](https://github.com/stonier/cost_map/pull/19) on that too looking to improve it a bit.

At any rate, it could very easily be templatised to handle `grid_map` and `cost_map` objects together I think.

skohlbr:
Thanks for the link @stonier! I made my own small converter by pulling out some code from navigation, converting from CostMap2D to nav_msgs/OccupancyGrid and then just using the conversion to GridMap. Not terribly efficient, but gets the job done: https://github.com/tu-darmstadt-ros-pkg/grid_map_navigation_planner/blob/master/grid_map_planner_nav_plugin/include/grid_map_planner_nav_plugin/grid_map_cost_map_2d_conversion.h#L71

stonier:
Oh, two step. Bit expensive, but it should work.

Good to have this issue here. I've been a bit delayed getting my cost map revamped (has to wait until I get back from interview travels), but this would be a good function to get into the core for both cost_map and grid_map.

pfankhauser:
Thanks @skohlbr and @stonier for your input! I added a quick implementation of a Costmap2D to Grid Map in the [feature/costmap_conversion](https://github.com/ethz-asl/grid_map/tree/feature/costmap_conversion) branch.

- The class `Costmap2DConverter` provides two methods
    1. First one has to initialize a new grid map with  `initializeFromCostmap2d` to adapt to the geometry of the costmap.
[...]

PR #78, opened 2016-11-24
Feature/costmap conversion

pfankhauser:
Adds a new conversion for Costmap2D types from ROS Navigation (#76).

### DIFF
--- a/grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
+++ b/grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
@@ -4,52 +4,137 @@
  *  Created on: Nov 23, 2016
  *      Author: Peter Fankhauser, ETH Zurich
  *              Stefan Kohlbrecher, TU Darmstadt
+ *              Daniel Stonier, Yujin Robot
  */
 
 #pragma once
 
+#include <grid_map_core/grid_map_core.hpp>
+
 // ROS
 #include <costmap_2d/costmap_2d.h>
 #include <costmap_2d/costmap_2d_ros.h>
+#include <tf/tf.h>
 #include <ros/ros.h>
 
 // STD
 #include <vector>
+//#include <sstream>
 
 namespace grid_map {
 
+/*!
+ * @brief Direct cost translations for costmap_2d -> grid/cost maps.
+ *
+ * This maps the cost directly, simply casting from the underlying
+ * unsigned char to whatever onto float fields between 0.0 and 100.0 with
+ * reserved values allocated for the costmap_2d.
+ */
+class Costmap2DDirectTranslationTable
+{
+ public:
+  Costmap2DDirectTranslationTable() {}
+
+  template<typename DataType>
+  void apply(std::vector<DataType>& costTranslationTable) const
+  {
+    costTranslationTable.resize(256);
+    for (unsigned int i = 0; i < costTranslationTable.size(); ++i) {
+      costTranslationTable[i] = static_cast<DataType>(i);
+    }
+  }
+};
+
+/**
+ * @brief Cost translations to [0, 100] for costmap_2d -> grid/cost maps.
+ *
+ * This maps the cost onto float fields between 0.0 and 100.0 with
+ * reserved values allocated for the costmap_2d.
+ */
+class Costmap2DCenturyTranslationTable
+{
+ public:
+  Costmap2DCenturyTranslationTable() {}
+
+  void apply(std::vector<float>& costTranslationTable) const
+  {
+    costTranslationTable.resize(256);
+    costTranslationTable[0] = 0.0;     // costmap_2d::FREE_SPACE;
+    costTranslationTable[253] = 99.0;  // costmap_2d::INSCRIBED_INFLATED_OBSTACLE;
+    costTranslationTable[254] = 100.0;  // costmap_2d::LETHAL_OBSTACLE
+    costTranslationTable[255] = -1.0;  // costmap_2d::NO_INFORMATION
+
+    // Regular cost values scale the range 1 to 252 (inclusive) to fit
+    // into 1 to 98 (inclusive).
+    for (int i = 1; i < 253; i++) {
+      costTranslationTable[i] = char(1 + (97 * (i - 1)) / 251);
+    }
+  }
+};
+
+template<typename DataType>
+class Costmap2DDefaultTranslationTable : public Costmap2DDirectTranslationTable {};
+
+template<>
+class Costmap2DDefaultTranslationTable<float> : public Costmap2DCenturyTranslationTable {};
+
+/*!
+ * Converter.
+ */
 template<typename MapType>
 class Costmap2DConverter
 {
- public:
+public:
+
+  /*!
+   * Initialises the cost translation table with the default policy for the template
+   * map type class.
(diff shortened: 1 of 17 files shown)


grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
```diff
@@ -0,0 +1,107 @@
...
+    initializeConversionTable();
+  }
+
+  virtual ~Costmap2DConverter()
+  {
+  }
+
+  void initializeFromCostmap2d(const costmap_2d::Costmap2D& costmap2d, MapType& outputMap)
+  {
+    Length length(costmap2d.getSizeInMetersX(), costmap2d.getSizeInMetersY());
```
stonier: From `costmap_2d.h`

```
  /**
   * @brief  Accessor for the y size of the costmap in meters
   * @return The y size of the costmap (returns the centerpoint of the last legal cell in the map)
   */
  double getSizeInMetersY() const;
```

stonier: Just noticed its covered in the next line...

pfankhauser:
I have merged this with the additions from #84 and will add some comments.

grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
```diff
@@ -0,0 +1,314 @@
...
+
+    // Get the Robot Pose Transform.
+    tf::Stamped<tf::Pose> tfPose;
+    if(!costmap2d.getRobotPose(tfPose))
+    {
+      std::ostringstream error_message;
+      error_message << "Could not get robot pose, is it actually published?";
+      ROS_ERROR_STREAM("CostmapConverter:: " << error_message.str());
+      return false;
+      // throw std::runtime_error(error_message.str());
```
pfankhauser: I'm happy to use exceptions instead of the `ROS_ERROR` and returning `false`. Is there a common sense when to use which?

grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
```diff
@@ -0,0 +1,314 @@
...
+      newCostMapOrigin(0) = std::floor(robotCellPosition.x())*resolution + resolution/2.0 + rosMapOrigin.x();
+    } else {
+      newCostMapOrigin(0) = std::round(robotCellPosition.x())*resolution + rosMapOrigin.x();
+    }
+    if ( number_of_cells_y % 2 ) { // odd
+      newCostMapOrigin(1) = std::floor(robotCellPosition.y())*resolution + resolution/2.0 + rosMapOrigin.y();
+    } else {
+      newCostMapOrigin(1) = std::round(robotCellPosition.y())*resolution + rosMapOrigin.y();
+    }
+
```
pfankhauser: I feel like this could be done easier, for example with getting the index for the robot position (`getIndexFromPosition`) and then using that index's position (`getPositionFromIndex`). In general, I'm in favor of moving these math related methods to `GridMapMath.hpp` for later re-use.

grid_map_ros/test/Costmap2DConverterTest.cpp
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  // Copy data.
+  costmap2dConverter.addLayerFromCostmap2D(costmap2d, layer, gridMap);
+
+  // Check data.
+  for (const auto& testValue : testValues) {
+    EXPECT_EQ(std::get<2>(testValue), gridMap.atPosition(layer, std::get<0>(testValue)));
+  }
+}
```
pfankhauser: @stonier if you have some tests to check the (so far) untested methods, I'm happy to include them also for Grid Map.

stonier: Could certainly be dropped in in a similar way here for the grid map test.

* [tests](https://github.com/stonier/cost_map/tree/devel/cost_map_demos/src/test/rostests/from_ros_costmaps)
* demo rviz application running similar code ([cpp](https://github.com/stonier/cost_map/blob/devel/cost_map_demos/src/applications/from_ros_costmaps.cpp) and [roslaunch](https://github.com/stonier/cost_map/blob/devel/cost_map_demos/launch/from_ros_costmaps.launch))
* library code supporting the tests and the demo ([hpp](https://github.com/stonier/cost_map/blob/devel/cost_map_demos/include/cost_map_demos/from_ros_costmaps.hpp) and [cpp](https://github.com/stonier/cost_map/blob/devel/cost_map_demos/src/lib/from_ros_costmaps.cpp))

grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
stonier: ROS_XXX in a library can be inconvenient to the user -> at best you take away their choice of how to log errors and at worst it can be a misdirection for other devs and test engineers.

In the former case, they may have a different logging system and so this just creates noise and loses the string message. In the latter case, they might actually handle the `false` return and recover from outside the function, at which point it is no longer logically an error.

Fundamentally, when you have more than one possible error source in a method, you generally need to give the user a notification that it went wrong as well as why it went wrong.

Exceptions do that via the throw, the type and the string message, but there are cases (e.g. high frequency call) where you don't want to burden it with throwing and catching every time you hit the method. This might be the case for cost maps.

In these cases I usually drop back to having the class return either true/false or an enum backed with a string `error_message` variable in the class which maintains the last known result string (bit like the posix `enum`/`strerror()` functionality). That is probably more appropriate than exceptions here since cost map calls could be high frequency.

Another interesting option I've seen recently is the `Exected<T>` template pattern. The idea originated in other languages (e.g. Haskell's `Maybe`). You can see some slides with code samples [here](https://onedrive.live.com/?cid=F1B8FF18A2AEC5C5&id=F1B8FF18A2AEC5C5%211158&parId=root&o=OneUp). Page 31 and 32 show how to use it to return appropriately in a function and how to handle it respectively. The actual `Expected<T>` implementation code is a bit more complicated and shown on the previous pages. 

Been wanting to play around with that, but in someone else's library is probably not the place to experiment :)

grid_map_ros/include/grid_map_ros/Costmap2DConverter.hpp
stonier: Aye, thought it should be simple as well. One of the other lads here shifted to index first, and then back to position, but had problems with numerical accuracies when sitting on the boundary. Shifting back and forth seems like overkill to me. Those functions don't entirely suit either - what we are seeking for here is the `mapPosition` variable, which in both of those methods is a known incoming variable of the method.

Not sure also it would be eminently reusable. We're seeking to align a grid map on top of an 'other' map that is as logically close as possible to a specified position. This needs information about the 'other' map (e.g. what is it's convention for the origin?). I'd leave this here and if we find another use case emerging, then do something about it.

stonier:
Test added and ros logging replaced by bool/error message strings.

@pfankhauser should be good to go now.

pfankhauser:
Looks great, thanks! I will merge this and create a new release asap.

stonier:
Ah, forgot to give it a quick run over to make it consistent. Thanks.

PR #83, opened 2016-12-28
Type traits for GridMap

stonier:
Useful for pulling the data and container types from the class if it has been passed as a template parameter to a template method/class.


stonier:
Could be more formally embedded in an external traits class, but since there is access to do it internally this is simpler.

stonier:
Dropping this in favour of having it embedded in #84 where it gets used.

PR #84, opened 2016-12-28
Costmap 2d Converter

stonier:
Started with the implementation in pull request #78 and extended it to accomodate:

* direct costmap2D -> cost_map translation (i.e. not 0.0 -> 100.0)
* conversions of subwindows around the costmap2DROS robot pose

Note that it also touched the core `GridMap` class to add some type traits that assist template methods work with either of the map types. I've a more complete set of `cost_map::CostMap` tests for this class in the `cost_map_ros` package.


stonier:
Would be nice to do the error and log handling a little differently - throw exceptions instead of `ROS_ERROR` style so the user then has the choice about how they want to handle it. Thoughts?

pfankhauser:
Thanks a lot @stonier for the great additions! I merged your changes with the original PR #78 to keep the flow of changes. For another time, you could also directly work on the PR of this repository since you have writing rights as a member ;-)

stonier:
Aye, got it. Closing this one down.

PR #86, opened 2017-01-20
re-implemented interpolation methods more efficiently

todorangrg:
Hello,

I have been using the interpolated version of atPosition and I noted (using gperftools) that in my code, 40% of the CPU time is spent in this function. So I re-implemented the interpolation in a more efficient way (it is basically reducing the coordinates to 4 points of coordinates (0,0), (1,0), (0,1),(1,1) ). Now atPosition uses 8% of CPU time in my application.


pfankhauser:
Dear @todorangrg, thanks a lot for sharing this. I'd like @djud to have a look before we merge.

djud:
Seems fine!

PR #92, opened 2017-03-22
Fixes transparency issue, where robots and other markers could not be…

### LINKED ISSUE #68: Grid Map RViz Plugin: Alpha Channel
Only other grid_map's can be seen if a grid_map is set to transparent (alpha<1). Robot model and other markers are not visible through transparent grid_map's.

grid_map_rviz_issue_1
grid_map_rviz_issue_2

pierrefranklin:
… seen through a transparent Grid_map.

fixes #68 

I'm not entirely sure what advantages/disadvantages setting DepthWrite gives you. However, changing it to false should fix the transparency issue.


pierrefranklin:
These commits should fix transparency with the text in rviz, and also add transparency to the grid lines.

pfankhauser:
Great, thanks @pierrefranklin!

PR #96, opened 2017-05-04
Feature/costmap 2d separation

pfankhauser:
I have separated the costmap_2d conversion functionality into a separate package (#94). This way, users working only with the standard installation of ROS can use `grid_map_ros` without installing `costmap_2d` first.

@stonier: Could you please check if this still works for your projects before I merge? Thanks!


stonier:
@pfankhauser no longer at Yujin, but I can certainly at least check it keeps working with `cost_map` and it's tests. Will check it on the weekend once I have some cycles.

stonier:
Not forgotten. Just a heads up to say sorry - had a guest arrive last weekend. Rescheduled for this weekend which I have free.

stonier:
Green light here...I have a branch to PR as soon as this goes in.

pfankhauser:
Great, thanks for checking!

PR #97, opened 2017-05-04
Feature/octomap conversion

pfankhauser:


pfankhauser:
Looks great, thanks a lot, @jeffdelmerico! I updated the code such that it also works under ROS Kinetic and Ubuntu 16.04. Could you please check if things still run fine for you?

PR #103, opened 2017-06-15
Feature/efficient mesh converter

alexmillane:
Makes the conversion from a mesh to a grid map more efficient. Makes many orders of magnitude difference when working with meshes produced from stereo which have 10 of thousands of triangles.


pfankhauser:
Thanks @alexmillane! @djud could you please look at this to confirm it still works with your projects?

alexmillane:
@pfankhauser @djud Any update on this?

PR #104, opened 2017-06-20
Fix/Remove constructor and destructor declaration in GridMapCvConverter

wsascha:
Having only the declarations yields to undefined references, so I removed them.


pfankhauser:
Thanks!

PR #105, opened 2017-06-23
Updated visualization to support triangles in all directions.

StefanFabian:
I've noticed that the grid map visualization only supports edges in one direction.
This pull request updates the visualization to support edges in all directions.
The following are example images from the current master state and the new visualization:

**Old**
grid_map_before

**New**
grid_map_after


StefanFabian:
Oh, I forgot.
I've removed the plottedFirstLine bool and just always open a new line after adding the two points.
It seems to work just fine that way but I don't know if there was a special reason to do it like that.

pfankhauser:
This looks great, thanks a lot! I will have a detailed look at it asap.

Issue #88, opened 2017-02-07
Alignment issues with Eigen datatypes and vectorization


mjschuster:
Hi there, 
first of all, thanks for this easy-to-use and well documented library!

There however seem to be alignment issues with the Eigen datatypes when compiling the library and software including its headers, which can lead to segmentation faults with compiling with vectorization (SSE) enabled. This probably is the case as you use Eigen datatypes like ```Eigen::Vector2d``` as a member of a structs/classes or in ```stl``` containers, like here:
https://github.com/ethz-asl/grid_map/blob/master/grid_map_core/include/grid_map_core/Polygon.hpp#L206

See https://eigen.tuxfamily.org/dox-devel/group__DenseMatrixManipulation__Alignement.html for more information on the particular issues. 

As far as I can see:
- best would be to adapt the code to use the Eigen allocators with correct alignment for vectorization 
- second best to use ```Eigen::DontAlign``` datatypes to prevent alignment issues
- third best to at least warn the user to turn off Eigen vectorization as otherwise it can lead to hard-to-debug segmentation faults
(see Eigen docs linked above for details on each of the three options)

pfankhauser:
Dear @mjschuster, thanks a lot! This might actually be an issue and I can look into this at a later point. If you have a suggestion on how to implement this fix, feel free to create a PR.

mjschuster:
Dear @pfankhauser, thanks! 
It'd be great if you could have a look at this. I myself probably won't have the time for a pull request on this issue anytime soon. My suggestion would be to go through the code and see if its possible to 
- use Eigen aligned allocators for all structs / classes having fixed-size vectorizable Eigen datatypes as members (see https://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html) (that should be easy)
- make sure not to pass Eigen objects by value (see https://eigen.tuxfamily.org/dox-devel/group__TopicPassingByValue.html)
- use Eigen aligned allocators for all stl containers with fixed-size alignable Eigen datatypes or classes containing them (see https://eigen.tuxfamily.org/dox-devel/group__TopicStlContainers.html)  (this might be the most difficult / annoying part) 

[...]

PR #110, opened 2017-07-12
Fix Eigen alignment in std::vector

thomasemter:
Addresses issue #88.

### DIFF
--- a/grid_map_core/include/grid_map_core/GridMap.hpp
+++ b/grid_map_core/include/grid_map_core/GridMap.hpp
@@ -349,7 +349,7 @@ class GridMap
    * @param newRegions the regions of the newly covered / previously uncovered regions of the buffer.
    * @return true if map has been moved, false otherwise.
    */
-  bool move(const Position& position, std::vector<BufferRegion>& newRegions);
+  bool move(const Position& position, std::vector<BufferRegion, Eigen::aligned_allocator<BufferRegion> >& newRegions);
 
   /*!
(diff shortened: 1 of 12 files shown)


grid_map_core/include/grid_map_core/GridMap.hpp
```diff
@@ -514,7 +514,7 @@ class GridMap
   Time timestamp_;
 
   //! Grid map data stored as layers of matrices.
-  std::unordered_map<std::string, Matrix> data_;
+  std::unordered_map<std::string, Matrix, std::hash<std::string>, std::equal_to<std::string>, Eigen::aligned_allocator<std::pair<std::string, Matrix> > > data_;
```
MirkoFerrati: @ThomasEmter I am trying to compile this branch with clang, and I am getting errors similar to https://github.com/JakobEngel/dso/issues/111
Following the same fixes in ^, I added `const` before `std::string` in the line above and passed clang compilation. You may want to change it here. 
I did not investigate the implications of using a const string except the fact that it compiles with clang. 
Gcc is fine in both ways.

ThomasEmter: @MirkoFerrati thank you for looking into the pull request. You are absolutely right, the `std::string` has to be be `const` according to [Eigen docs](https://eigen.tuxfamily.org/dox/group__TopicStlContainers.htm). I must have missed that and I indeed only used GCC, so I did not get the compile errors.
Should be fixed now.

PR #114, opened 2017-07-26
Fix segfault in SpiralIterator the center is outside the map

2scholz:
In case the center of the `SpiralIterator` happens to be outside the map
a segmentation fault occurs when trying to use the iterator.

In this case the index of the center was not computed,
because `getIndexFromPosition` returned false without changing
the value of the index. With this patch, the index is computed, but the function still
returns false when the position is outside the map.

Additionally, only the cells adjacent to the center were looked at in that case.
When these cells are outside the map too, dereferencing the pointer leads
to a segmentation fault. This was fixed by generating rings until a position inside the map is reached.


v4hn:
Thank you a lot for debugging this @2scholz !

@pfankhauser a student of ours hit this bug last semester when he tried to look at the surrounding of points from a laser scan (apparently outside the map...).

PR #115, opened 2017-07-27
Fixed: Clamp changes were disregarded

Le-Fix:
Bug has been noticed after converting a grid map and clamping the ray count to a reasonable number.
Converting the non clamped values produces the invalid picture.

grid_map_count_max255_post_fix
grid_map_count_max255_pre_fix


pfankhauser:
Great, thanks!

PR #117, opened 2017-08-03
grid_map_core: Polygon triangulation checks number of vertices.

remod:
I found this potential issue while debugging [https://github.com/leggedrobotics/free_gait/pull/39](url).
This is only a suggestion and not needed to solve the free gait visualization issue.


pfankhauser:
👍

PR #121, opened 2017-08-17
Feature/signed distance field

mktk1117:
- Made a Class that computes Signed Distance Field(SDF) from grid map efficiently (O(n^2)).
- To compute SDF use `computeSignedDistanceField` function. After that, the SDF is stored.
- To access the SDF, use `getDistanceAt` function. This returns the signed distance at the position.
- Also, `getDistanceGradientAt` provides the distance gradient at the position.


pfankhauser:
Closing this in favor if #122.

PR #126, opened 2017-08-29
Feature/signed distance field

mktk1117:
Made a namespace distance_transform


mktk1117:
I modified the gradient calculation.

PR #128, opened 2017-09-05
explicit initialization of member variables

ompugao:
in some case, default initialization does not work somehow and isPastEnd() returns false. the iterator returns a far away index.

### DIFF
--- a/grid_map_core/include/grid_map_core/iterators/LineIterator.hpp
+++ b/grid_map_core/include/grid_map_core/iterators/LineIterator.hpp
@@ -108,10 +108,10 @@ private:
   Index end_;
 
   //! Current cell number.
-  unsigned int iCell_;
+  unsigned int iCell_ = 0;
 
   //! Number of cells in the line.
-  unsigned int nCells_;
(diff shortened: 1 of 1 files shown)


AlexReimann:
Bump, have the same problem.
This happens if the line is outside your map boundaries.
`intitialize()` is not called in the constructor because `getIndexLimitedToMapRange` returns `false`:

https://github.com/ANYbotics/grid_map/blob/c0d7f0e80d4995fe75bff3e49ec7d4d9e150ec98/grid_map_core/src/iterators/LineIterator.cpp#L16-L23

Might as well always call `initialize` or have a flag to check if the iterator was properly initialized

AlexReimann:
Bump, would be nice if we don't keep having to build our own packages just because of this.

maximilianwulf:
@hogabrie, can you import that one?

AlexReimann:
Bump. Missed in the latest release again.

maximilianwulf:
Yes, we are sorry.

maximilianwulf:
Closed in favor of #224

PR #133, opened 2017-10-01
Increased robustness in cases where map is empty when computing SDF.

pfankhauser:


grid_map_sdf/src/SignedDistanceField.cpp
```diff
@@ -35,17 +36,17 @@ void SignedDistanceField::calculateSignedDistanceField(const GridMap& gridMap, c
   resolution_ = gridMap.getResolution();
   position_ = gridMap.getPosition();
   size_ = gridMap.getSize();
-  Matrix map = gridMap.get(layer);
+  Matrix map = gridMap.get(layer); // Copy!
 
   // Get min and max height and fill the max height in the NAN area.
```
mbjelonic: Minor mistake in the comment: Currently it is set to the lowest hight (i.e. -1e5) and not max height.

PR #134, opened 2017-10-06
Check position before computing index, add const's

wsascha:


pfankhauser:
Thanks! I'll check why the tests fail when I get around to it.

PR #138, opened 2017-11-05
Feature/more iterators

pierrefranklin:
Adds the 
Fill Iterator
Nearest Valid Iterator
Thickened Line Iterator (LineThickIterator)


AndyHuska:
+1 for pulling this PR.  The line drawing method in this branch is much more convenient than the basic line drawing in master because it does its own bounds checking.  Thick line drawing is very helpful as it provides a faster way to effectively draw large polygons, it also has an advantage over the basic polygon drawing it includes the polygon edge rather than just filling up to the edge which causes edges to get lost in large scale maps (e.g. 1m resolution).

I've been using this branch for a few weeks to add complex linestrips and polygons to very large GridMaps.  Other than a simple CMakeFile merge there is nothing preventing it from being used in production for the LineThickIterator which leverages the added FillIterator.  The NearestValidIterator and SpiralIterator I have not used, but they function in the demo.

PR #140, opened 2017-11-12
Adds in dependencies on OpenCV. 

pierrefranklin:
Previously all needed components were included by cv_bridge, but now cv_bridge does not include all opencv components, so they need to be explicitly added in other packages.

Specifically, about 3 months ago cv_bridge was updated to use
```
find_package(OpenCV 3 REQUIRED
COMPONENTS
    opencv_core
    opencv_imgproc
    opencv_imgcodecs
  CONFIG
)
```
instead of 
`find_package(OpenCV REQUIRED)`

https://github.com/ros-perception/vision_opencv/commit/8b5bbcbc1ce65734dc600695487909e0c67c1033

Previously, all of the components of OpenCV were included in the dependent libraries because of this; however, now only the listed libraries are there. So master failed to compile because the highgui library is not included. Doing another find_package in the dependent libraries lets you add in the missing (highgui) libraries.

### DIFF
--- a/grid_map_cv/CMakeLists.txt
+++ b/grid_map_cv/CMakeLists.txt
@@ -10,6 +10,8 @@ find_package(catkin REQUIRED COMPONENTS
   filters
 )
 
+find_package(OpenCV 3 REQUIRED)
+
 ###################################
 ## catkin specific configuration ##
 ###################################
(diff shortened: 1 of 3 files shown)


pierrefranklin:
@pfankhauser could you look at these build logs and tell me if its something I need to fix? It seems like an issue on the build server, but I don't understand the errors well enough to say for sure.

Regardless, this latest commit should fix the rosbag test, based on testing on a different pull request.

pierrefranklin:
Nevermind. Apparently the build magically works now.

PR #144, opened 2017-11-21
Resolve ignored qualifiers

wsascha:
Resolve for ignored qualifiers that appear when built with `-Wignored-qualifiers` option

### DIFF
--- a/grid_map_core/include/grid_map_core/Polygon.hpp
+++ b/grid_map_core/include/grid_map_core/Polygon.hpp
@@ -84,7 +84,7 @@ class Polygon
    * Returns the number of vertices.
    * @return the number of vertices.
    */
-  const size_t nVertices() const;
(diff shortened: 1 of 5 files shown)


romainreignier:
Nice PR because I am bored of the warnings.
But you should modify the cpp files also, like [here](https://github.com/ethz-asl/grid_map/blob/b0dedba838141b3de2d4d008c8fcfddc740c6f4d/grid_map_core/src/Polygon.cpp#L70) for the build to success.

wsascha:
Upsi, sorry from my side. I was seeing the warnings when building with gcc 7 and thought I quickly apply changes using the web interface without building the package locally... I guess that's considered bad practice :wink:

Implementations are adapted and I fixed another `-Wmaybe-uninitialized` warning without making another PR.

PR #146, opened 2017-11-21
Possible master compilation fix (OpenCV lib issue)

pierrefranklin:
Removes the components part of the grid_map_demos opencv find_package call


pierrefranklin:
@pfankhauser I somewhat doubt this will fix the master build, but I think it's worth a shot.
It just takes out the Components part of the find_package call.

Issue #148, opened 2017-11-29
[grid_map_core] Determine length of line within grid map cell


wsascha:
**Why?**
For occupancy mapping, I want to use a decay rate model, e.g. as presented [here](http://ais.informatik.uni-freiburg.de/publications/papers/schaefer17icra.pdf). Therefore, I need the distance a ray travelled within each cell.

**How?**
[This](https://stackoverflow.com/questions/4248090/finding-the-length-of-a-ray-within-a-cube) thread describes how that length can be computed for general convex polyhedrons. However, we have the very simple case of axis aligned 2D squared cells.

So I started an implementation (below) that currently needs an [`Eigen::ParameterizedLine`](https://eigen.tuxfamily.org/dox/classEigen_1_1ParametrizedLine.html) and the cell position. As this must be explicitly calculated I am wondering on a way on how to do that more efficiently.

Any suggestions? Anyone familiar with all the index handling?

**Reference impl** (I will create a linked PR soon)

A gist can be found [here](https://gist.github.com/wsascha/a08faa7b1065df70701a8b140b4aa8c2).
Note the `length()` method with its impl and adaptations in the `initialize()` method.

PR #149, opened 2017-11-30
[feature] [grid_map_core] Add length computation in LineIterator

wsascha:
See #148


PPokorski:
What about this PR? I would really like to see it merged and added to API. However, @wsascha, it would be nice to see some unit tests for it.

wsascha:
Ok, I'll implement one or two tests.

PR #152, opened 2017-12-05
cmake fix typo eigen include dirS

artivis:
Fix typo in CmakeLists related to eigen include dir variable.

Not sure why but on my machine `EIGEN3_INCLUDE_DIR` is empty while `EIGEN3_INCLUDE_DIRS` has the include path properly defined.

Ubuntu 14.04 + ROS Indigo.


artivis:
Looks like the inverse problem happens on Jenkins.  
I'll close this PR in favor of an issue...

PR #172, opened 2018-06-26
Feature/clear sub map

djud:
I added a service to clear only some regions of a grid map. I use this in elevation mapping to clear regions of the map that I know changed through interaction of the robot, e.g. by digging off a certain part or dumping soil with the excavator. See the corresponding [PR in elevation_mapping](https://github.com/ethz-asl/elevation_mapping/pull/73). The map will thus adapt faster and more reliable to changes.


grid_map_core/src/GridMap.cpp
```diff
@@ -633,6 +633,28 @@ void GridMap::clearAll()
   }
 }
 
+void GridMap::clearSubMap(const GridMap& subMap)
+{
+  const grid_map::Matrix& mapClear = subMap["clear"];
+  for (auto& data : data_) {
+    auto layerIt = std::find(subMap.getLayers().begin(),subMap.getLayers().end(),data.first);
```
pfankhauser:
```suggestion
    auto layerIterator = std::find(subMap.getLayers().begin(), subMap.getLayers().end(), data.first);
```

grid_map_core/src/GridMap.cpp
```diff
@@ -633,6 +633,28 @@ void GridMap::clearAll()
   }
 }
 
+void GridMap::clearSubMap(const GridMap& subMap)
```
pfankhauser: `Submap`, not `SubMap`

grid_map_core/src/GridMap.cpp
```diff
@@ -633,6 +633,28 @@ void GridMap::clearAll()
...
+{
+  const grid_map::Matrix& mapClear = subMap["clear"];
```
pfankhauser: What happens if layer "clear" does not exist?

grid_map_core/include/grid_map_core/GridMap.hpp
```diff
@@ -381,6 +381,12 @@ class GridMap
    */
   void clearAll();
 
+  /*!
+  * Clears cells that are set to 1.0 in the layer "clear".
```
pfankhauser: According to the code below, the cells are not only cleared but overwritten with the values from the new submap. Can you describe this?

grid_map_msgs/srv/ClearSubGridMap.srv
```diff
@@ -0,0 +1,4 @@
+# Sub map
```
pfankhauser: I think there's some confusion about the term "submap". So far I used it to describe a map which lies inside the original map but is smaller. In your case, the maps are of the same size, so submap does not apply.

My suggestions is to not use submap, but call rename the service to "ProcessMap.srv". The topic in Elevation Mapping can then be more specific.

djud:
@pfankhauser I have completely refactored my changes. Your comments were correct. I have changed the service to clear certain indices in the map. I think this is a much clearer interface and it should be clear what it is and does. I removed the possibility of reinitializing cleared values and handle this in my code. It is nicer that way. Is this better? If yes, I'll write the unit tests.

PR #174, opened 2018-07-07
Feature - Grow!

### LINKED ISSUE #132: No possibility to `resize` a GridMap ?
Hi,
I started using the `grid_map_core` pkg recently aiming at incrementally building a `GridMap`. However it seems there are no possibility to resize 'on the fly'  a `GridMap` (resize while preserving the existing data).   
The `setGeometry` functions internally call `clearAll` thus aren't suited for incremental building. I came across both `addDataFrom` & `extendToInclude` functions which both require as input argument another `GridMap` which is a bit of an overshoot - instantiation of a second `GridMap` object, internal copy of the current object etc...  
Did I miss something or is it a missing functionality ??

raghavakumar8:
Adds the ability to grow a gridmap instance efficiently (using Eigen functions) in different directions.

Resolves #132.

### DIFF
--- a/grid_map_core/src/GridMap.cpp
+++ b/grid_map_core/src/GridMap.cpp
@@ -63,7 +63,9 @@ void GridMap::setGeometry(const Length& length, const double resolution,
 
   resolution_ = resolution;
   length_ = (size_.cast<double>() * resolution_).matrix();
+
   position_ = position;
+
   startIndex_.setZero();
 
   return;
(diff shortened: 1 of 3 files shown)


grid_map_core/src/GridMap.cpp
```diff
@@ -717,5 +762,62 @@ void GridMap::resize(const Index& size)
...
 
+void GridMap::conservativeResize(const Index& size, const Direction direction, float value)
+{
+  int row_delta = size(0) - size_(0);
+  int col_delta = size(1) - size_(1);
+
+  Eigen::MatrixXf like = Eigen::MatrixXf::Constant(size(0), size(1), value);
+  for (auto& data : data_)
+  {
+    data.second.conservativeResizeLike(like);
```
artivis: Plain curiosity,  
why are you using `conservativeResizeLike` rather than `conservativeResize` ??  
The function `conservativeResize` allocates a temporary matrix under the hood, so I would imagine that `conservativeResizeLike` does the same. Thus isn't the temporary matrix `like` a waste of memory ?

raghavakumar8: `conservativeResize` doesn't let you initialize the values of new elements, `conservativeResizeLike` does. This is important when you want to be able to guarantee that every new grid cell is initalized to 0.0, for instance.

In principle I could use `conservativeResize`, loop through all new cells and manually initialize them, but I suspect that would be less efficient.

artivis: Ok I see your point.
Still I think a plain `conservativeResize` followed by manually setting the uninitialized values would be better as you don't really need allocating matrix `like`, especially as the matrix grows large.  
Rather than looping over uninitialized values you can use the `block` functions, something along the lines :
```cpp
data.second.conservativeResize(size(0), size(1));
// Assuming that resizing to larger dimension appends 
// rows at the bottom and cols at the right side.
// Set value for appended bottom rows
data.second.block(size_(0), 0, row_delta, size(1)).setConstant(value); 
// Set value for appended top-left block (remaining unintialized)
data.second.topRightCorner(size(0), col_delta).setConstant(value);
```
I haven't checked that the indices are actually correct ^^, but that's the general idea.

BrunoEduardoCSantos: Is this feature working? What is it stopping from going to production?

raghavakumar8: Just waiting for an approving review. We have been using this feature extensively in production at our [company](https://github.com/Maidbot/).

rupertbayern: this is very nice feature and exactly what I needed! A merge would be good.

SteveQUT: +1 for useful feature and would be good to have merged

maximilianwulf: Yes, will import it.

maximilianwulf:
@raghavakumar8 could you merge master back into your branch?

maximilianwulf:
Closing in favor of #330 which has more up-to-date master.

PR #197, opened 2019-01-22
Feature/process bag srv

marco-tranzatto:
Functions [`saveToBag`](https://github.com/ANYbotics/grid_map/blob/master/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp#L237) and [`loadFromBag`](https://github.com/ANYbotics/grid_map/blob/master/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp#L247) from "GridMapRosConverter.hpp" require an absolute path of the rosbg and a topic name.
Service ['ProcessFile.srv'](https://github.com/ANYbotics/grid_map/blob/master/grid_map_msgs/srv/ProcessFile.srv) provides only the filed `file_path`, therefore a new service has been created to allow the specification of a topic name.

### DIFF
--- a/grid_map_msgs/CMakeLists.txt
+++ b/grid_map_msgs/CMakeLists.txt
@@ -24,6 +24,7 @@ add_service_files(
   FILES
   GetGridMap.srv
   GetGridMapInfo.srv
+  ProcessBag.srv
   ProcessFile.srv
 )
 
@@ -53,4 +54,3 @@ catkin_package(
         std_msgs
     #DEPENDS
 )
-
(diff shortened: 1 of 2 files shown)


marco-tranzatto:
ping @pfankhauser

pfankhauser:
The `ProcessFile` service is used [here](https://github.com/ANYbotics/elevation_mapping/blob/634bed3ae22c707f05d0ed6506562d7eb839736c/elevation_mapping/src/ElevationMapping.cpp#L444) to actually process ROS bags. Hence, adding another service type `ProcessBag` is misleading. Instead of creating a new service type, I suggest you extend the existing one with an array of strings with the description, that additional information (e.g. ros topics) can be stored there.

PR #198, opened 2019-01-24
Extend ProcessFile.srv to include topic_name field.

marco-tranzatto:
Functions [`saveToBag`](https://github.com/ANYbotics/grid_map/blob/master/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp#L237) and [`loadFromBag`](https://github.com/ANYbotics/grid_map/blob/master/grid_map_ros/include/grid_map_ros/GridMapRosConverter.hpp#L247) from "GridMapRosConverter.hpp" require an absolute path of the rosbg and a topic name.
Service ['ProcessFile.srv'](https://github.com/ANYbotics/grid_map/blob/master/grid_map_msgs/srv/ProcessFile.srv) is extended to allow the specification of a topic name.

### DIFF
--- a/grid_map_msgs/srv/ProcessFile.srv
+++ b/grid_map_msgs/srv/ProcessFile.srv
@@ -1,8 +1,8 @@
 # Absolute file path.
 string file_path
 
-# Topic name.
-string topic_name
+# For ROS bags: A list of topic names that should be processed (optional).
+string[] topic_names
 
 ---
 
--- a/grid_map_msgs/srv/ProcessFile.srv
+++ b/grid_map_msgs/srv/ProcessFile.srv
@@ -1,6 +1,9 @@
 # Absolute file path.
 string file_path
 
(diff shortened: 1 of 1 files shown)


pfankhauser:
What do you think about my change proposal?

marco-tranzatto:
@pfankhauser I actually prefer only one field, because I think it's easier to handle and in any case in each rosbag you want to process only one topic per service call. I believe that if you start adding the possibilities of processing multiple topics with one service call, then the user should be aware of the order of the topic names. For example topic_name#1 is used for XX whereas topic#2 is used for YY and so on. I think this complicates things, and we should aim at having different names of service calls, one for each topic we want to process. Therefore I'd have only one string instead of the array.

Do you see any practical example where you want multiple topic names?

marco-tranzatto:
single `topic_name` field restored, as agreed with @pfankhauser

PR #203, opened 2019-02-27
fix job status in readme, remove EOL Jade and add Lunar instead

samuelba:


remod:
Please add the new packages as well, thanks!

samuelba:
@remod Done.

PR #221, opened 2019-12-13
Fix potential crash of rviz plugin in case of NAN values within the grid

j-kallwies:
We have observed sporadic crashes of rviz, when we are using grid maps containing some NAN values.

The error message is:
```rviz: /build/ogre-1.9-B6QkmW/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreAxisAlignedBox.h:252: void Ogre::AxisAlignedBox::setExtents(const Ogre::Vector3&, const Ogre::Vector3&): Assertion `(min.x <= max.x && min.y <= max.y && min.z <= max.z) && "The minimum corner of the box must be less than or equal to maximum corner"' failed.```

This simple change fixes this problem.


maximilianwulf:
Thanks a lot! @hogabrie can you merge this one?

PR #231, opened 2020-06-04
Updates for noetic (message change)

jspricke:


jspricke:
Moving the header there allows to use message filters and time
synchronizers and unifies it with nav_msgs/OccupancyGrid.

grid_map_msgs/msg/GridMap.msg
```diff
@@ -1,4 +1,7 @@
-# Grid map header
+# Header (time and frame)
```
maximilianwulf, line 1: I have to check if that does not break our internal API.

jspricke:
>Thanks a lot for the submission. Does this MR provide the necessary changes to be working in noetic?

Yes, unit tests and demos all work.

>> @@ -1,4 +1,7 @@
>-# Grid map header
>+# Header (time and frame)
>
>I have to check if that does not break our internal API.

The Noetic release is the perfect (and last) opportunity to fix this. 
Also this simplifies the rviz plugin a lot. So it would be really great 
to see this merged.

remod:
Which part of the changes in this PR are minimally required to port grid map to noetic? I guess it is only switching from C++ 11 to 14, right?

Keep in mind that changing the message definition will break the people's code using this message, as well as recorded bag file datasets. Would it make sense to rethink the message when switching to ROS2, as this will most probably result other interface changes?

jspricke:
>Which part of the changes in this PR are minimally required to port grid map to noetic? I guess it is only switching from C++ 11 to 14, right?

Depends on the package, the PR is minimal. grid_map_sdf needs the C++14, 
grid_map_rviz_plugin needs more work.

>Keep in mind that changing the message definition will break the people's code using this message, as well as recorded bag file datasets. Would it make sense to rethink the message when switching to ROS2, as this will most probably result other interface changes?

Breaking messages between ROS releases (like melodic -> noetic) is fine 
and a common process. Note that the ROS wiki lists no users of these 
packages so it's hard to speculate what would break. If you have private 
APIs depending on it it would be great if you could share some details. 
For bag files you can add migration rules.

maximilianwulf:
Hey @jspricke, unfortunately, we are negative about the message change. But we are happy to change this for ROS2. 

We will upgrade the ROS2 branch very soon, can you make a PR of the message change to the ROS2 branch?

The other changes then can be integrated. 

Thanks a lot.

jspricke:
>Hey @jspricke, unfortunately, we are negative about the message change.

Can you give some reason for this?
Note that I would propose to do this change only in noetic, not im 
melodic. This is usually done by having a melodic/noetic branch.

>But we are happy to change this for ROS2.
>We will upgrade the ROS2 branch very soon, can you make a PR of the message change to the ROS2 branch?

I would welcome such a change but will not do it myself as I have not 
worked with ROS2 for years.

Cheers Jochen

maximilianwulf:
> Can you give some reason for this?

The main argument for this is the simplified RVIZ plugin, right?

maximilianwulf:
> Can you give some reason for this?

The general reason for that is a major API break for a lot of people. We think that the community accepted the rather weird msg definition and we don't want to enforce a change in ROS1. 

Do you see other major benefits of changing the msg definition beside the simplified RVIZ plugin?

I added a note in the ROS2 port PR #232 to update the message.

jspricke:
>The general reason for that is a major API break for a lot of people. We think that the community accepted the rather weird msg definition and we don't want to enforce a change in ROS1.

The change is rather minimal from my point of view.

>Do you see other major benefits of changing the msg definition beside the simplified RVIZ plugin?

Allowing to use message filters, time synchronizers and probably more 
ROS infrastructure depending on well defined messages.

maximilianwulf:
> The change is rather minimal from my point of view.

Unfortunately, I cannot disclose our internal usage, but it would require quite an effort for us to adapt. The benefits at this point do not outweigh the extra effort. 

Therefore for the master branch, I can only accept non API breaking changes. 

So we see two options:

- At the moment we only maintain the master branch, but if you want I can create a `noetic` branch and you add the changes there. 
- You keep the `header` in `GridMapInfo.msg` and we can check internally if it is back compatible.

jspricke:
* Maximilian Wulf <notifications@github.com> [2020-06-10 01:42]:
>- At the moment we only maintain the master branch, but if you want I can create a `noetic` branch and you add the changes there.

That would be awesome :).

maximilianwulf:
> That would be awesome :).

Okay, I think that is doable, but we won't actively maintain the noetic branch, could you have an eye on it?

jspricke:
* Maximilian Wulf <notifications@github.com> [2020-06-11 01:42]:
>Okay, I think that is doable, but we won't actively maintain the noetic branch, could you have an eye on it?

Would you do noetic releases and push them here?

https://github.com/anybotics/grid_map-release.git

jspricke:
Any news here? Would be great to get this released for Noetic.

UniBwTAS:
We also would be very happy when there will be a Release for Noetic of this repository.

maximilianwulf:
I am sorry for the delay, we are adjusting some things internally. We will make grid_map available for Noetic.

matthewhannay567:
Still waiting for an update on this.

maximilianwulf:
We are still working on it, but the Noetic release will come soon.

maximilianwulf:
Closing this because the necessary changes for Noetic have been introduced. 

@jspricke thank you again for the proposal :)

matthewhannay567:
When will grid_map be available as a noetic package?

maximilianwulf:
Hi @matthewhannay567, I cannot give a concrete date, but it will come.

Issue #219, opened 2019-12-08
ROS2 Port


jacobhuesman:
I'm going to start the process of porting `grid_map` over to ROS2. I noticed some chatter about it on #178, but haven't seen any branches or forks.

@pfankhauser can you create a `ros2` branch that I can submit pull requests against?

maximilianwulf:
Hey @jacobhuesman, PRs are very welcome!

Feel free to add PRs to this [ros2 branch](https://github.com/ANYbotics/grid_map/tree/feature/ros2).

SteveMacenski:
@maximilianwulf I have some folks working with me on navigation2-land that have interest in working to port over code to ROS2 with the goal of having this replace potentially costmap 2D (or chunks of it).

It might be good to have you join our community slack (navigation2.slack.com) so we can chat about this. I also want to figure out what a reasonable maintenance arrangement looks like if you're not currently using grid_maps in ROS2 (e.g. if PRs come in for 1 version syncing to both, if we need to get things into the ROS2 branch, are there maintainers willing to review and iterate, would it be possible to add me and/or Tully as reviewers strictly for ROS2, etc). 

Alternatively, we could work off of a fork of grid_maps in `ros-perception`, but I'd really like to keep the code with the owners when possible. Deep forks hurts everyone.

maximilianwulf:
Hi @SteveMacenski, thanks a lot for the request. We are very interested in helping to port `grid_map` to ROS2. 
Does that mean you already have a plan for the port and we should discuss it on slack?

> are there maintainers willing to review and iterate

Yes, we are willing to help review PRs.

> would it be possible to add me and/or Tully as reviewers strictly for ROS2

Yes, definitely. 

> Alternatively, we could work off of a fork of grid_maps in ros-perception, but I'd really like to keep the code with the owners when possible. Deep forks hurts everyone.

Agreed, we think the same and would like to have it on this repository.

SteveMacenski:
[...]

PR #233, opened 2020-06-05
grid_map_core ROS2 port

Marwan99:
Related to #219

Port of grid_map_core package to ROS2.

- source and tests linted
- CMakeLists updated
- package.xml updated


SteveMacenski:
Because this one is so large, can you highlight what you did? I imagine you updated the package xml and cmakelists then ran the ament_uncrustify automatic reformatter. What else? That will help me gloss over the auto-generated code changes and just focus in on the human changes

Marwan99:
Major hand made changes:
- Swapping `int` with `std::size_t`  in for loops " looping variable" in loops where the compiler complained about comparing signed to unsigned.
- Resolving namespaces by hand.
- replacing `#pragma once` header guards, used the names suggested by the linter.

Will highlight the others in a review below for convince

grid_map_core/src/GridMap.cpp
```diff
@@ -153,116 +173,137 @@ bool GridMap::erase(const std::string& layer) {
...
-      else
-        interpolationMethod = InterpolationMethods::INTER_NEAREST;
-    }
+        float value;
+        if (atPositionLinearInterpolated(layer, position, value)) {
+          return value;
+        } else {
+          interpolationMethod = InterpolationMethods::INTER_NEAREST;
+        }
+        [[fallthrough]];
```
Marwan99, line 224: using [fallthrough](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) (C++17, is that okay?) I think the logic relies on not having break statements at the end of the cases, but the compiler was not happy with with that.

grid_map_core/src/GridMapMath.cpp
```diff
@@ -463,105 +552,122 @@ bool incrementIndexForSubmap(Index& submapIndex, Index& index, const Index& subm
...
 
-void colorVectorToValue(const Eigen::Vector3i& colorVector, float& colorValue)
+void colorVectorToValue(const Eigen::Vector3i & colorVector, float & colorValue)
 {
-  unsigned long color = (colorVector(0) << 16) + (colorVector(1) << 8) + colorVector(2);
-  // cppcheck-suppress invalidPointerCast
-  colorValue = *reinterpret_cast<float*>(&color);
+  uint64_t color = (colorVector(0) << 16) + (colorVector(1) << 8) + colorVector(2);
+  colorValue = static_cast<float>(color);
 }
```
Marwan99, line 665: compiler was complaining about `*reinterpret_cast<float*>` replaced it with `static_cast<float>`

grid_map_core/src/Polygon.cpp
```diff
@@ -6,49 +6,52 @@
...
 {
   int cross = 0;
-  for (int i = 0, j = vertices_.size() - 1; i < vertices_.size(); j = i++) {
-    if ( ((vertices_[i].y() > point.y()) != (vertices_[j].y() > point.y()))
-           && (point.x() < (vertices_[j].x() - vertices_[i].x()) * (point.y() - vertices_[i].y()) /
-            (vertices_[j].y() - vertices_[i].y()) + vertices_[i].x()) )
+  for (std::size_t i = 0, j = vertices_.size() - 1; i < vertices_.size(); j = i++) {
+    if ( ((vertices_[i].y() > point.y()) != (vertices_[j].y() > point.y())) &&
+      (point.x() < (vertices_[j].x() - vertices_[i].x()) * (point.y() - vertices_[i].y()) /
+      (vertices_[j].y() - vertices_[i].y()) + vertices_[i].x()) )
```
Marwan99, line 41: Example of replacing `int` with `std::size_t` in a for loop

grid_map_core/src/GridMap.cpp
```diff
@@ -695,8 +813,9 @@ void GridMap::convertToDefaultStartIndex() {
   startIndex_.setZero();
 }
 
-Position GridMap::getClosestPositionInMap(const Position& position) const {
-  if (getSize().x() < 1u || getSize().y() < 1u) {
+Position GridMap::getClosestPositionInMap(const Position & position) const
+{
+  if (getSize().x() < 1 || getSize().y() < 1) {
```
Marwan99, line 818: Example of resolving a complain about comparing signed to unsigned.

grid_map_core/test/GridMapIteratorTest.cpp
```diff
@@ -21,43 +18,48 @@
...
+  auto & data = map["layer"];
   unsigned int i = 0;
   for (; !iterator.isPastEnd(); ++iterator, ++i) {
     data(iterator.getLinearIndex()) = 1.0;
     EXPECT_EQ(i, iterator.getLinearIndex());
     EXPECT_FALSE(iterator.isPastEnd());
   }
 
-  EXPECT_EQ(40, i);
+  EXPECT_EQ(40u, i);
```
Marwan99, line 62: Another example of resolving compiler complains about comparing signed to unsigned. This error was very common in the test files.

Marwan99:
@SteveMacenski If there is something else I can do to make it slightly easier please let me know. Apologies for the big PR.

maximilianwulf:
@SteveMacenski and @Marwan99 I think there are many improvements on the source that would also help in the ROS1 version. With newer packages, we enforce clang, which is not added. Would it make sense to update the ROS1 version first and then integrate it into ROS2?

YoshuaNava:
@maximilianwulf Does it make sense to split this PR into small ones?

grid_map_core/src/GridMap.cpp
SteveMacenski: You should use C++14 unless absolutely required to use C++17 - ROS2 Foxy is under C++14.

grid_map_core/src/GridMapMath.cpp
SteveMacenski: Fine by me :+1:

SteveMacenski:
@maximilianwulf most of these updates were because of the picky compiler flags we set in the common packages and because of the ROS2 linters. If things are working in ROS1, I wouldn't go back and try to make all these changes since it would be time consuming and for little tangible benefit. Something like a C style cast is totally valid, but the ROS2 style guide complains about it because it would rather use a std cast. 

@YoshuaNava for this one, potentially, but they'll all be massive. >50% of this work is done from automated linter reformatting.  If we were to split it up, it would make sense to split it up into 2 PRs
- Automated linting - just autoapprove assuming compiles and passes tests
- Hand linting - the things we would need to spend a little time just sanity checking.

grid_map_core/test/EllipseIteratorTest.cpp
```diff
@@ -21,16 +18,19 @@
 // Vector
 #include <vector>
 
-using namespace std;
-using namespace Eigen;
-using namespace grid_map;
+#include "grid_map_core/iterators/EllipseIterator.hpp"
+#include "grid_map_core/GridMap.hpp"
+
+// using namespace std;
```
maximilianwulf, line 24: Delete.

grid_map_core/test/GridMapIteratorTest.cpp
```diff
@@ -21,43 +18,48 @@
 // Vector
 #include <vector>
 
-using namespace std;
-using namespace grid_map;
+#include "grid_map_core/iterators/GridMapIterator.hpp"
+#include "grid_map_core/GridMap.hpp"
+
+// using namespace std;
```
maximilianwulf, line 24: Delete

grid_map_core/test/GridMapMathTest.cpp
```diff
@@ -20,231 +18,380 @@
 // Vector
 #include <vector>
 
-using namespace std;
-using namespace grid_map;
+#include <limits>
+
+#include "grid_map_core/GridMapMath.hpp"
+
+// using namespace std;
```
maximilianwulf, line 25: Delte

grid_map_core/test/LineIteratorTest.cpp
```diff
@@ -6,24 +6,27 @@
...
 #include <gtest/gtest.h>
 
 // Limits
 #include <cfloat>
 
-using namespace grid_map;
+#include "grid_map_core/iterators/LineIterator.hpp"
+#include "grid_map_core/GridMap.hpp"
+
+// using namespace grid_m ap;
```
maximilianwulf, line 18: Delete

grid_map_core/test/PolygonIteratorTest.cpp
```diff
@@ -6,40 +6,43 @@
...
-using namespace std;
-using namespace Eigen;
-using namespace grid_map;
+#include <string>
+
+#include "grid_map_core/iterators/PolygonIterator.hpp"
+#include "grid_map_core/GridMap.hpp"
+#include "grid_map_core/Polygon.hpp"
+
+// using namespace std;
```
maximilianwulf, line 27: Delete

grid_map_core/test/PolygonTest.cpp
```diff
@@ -6,59 +6,61 @@
...
 #include <Eigen/Dense>
 
-using namespace std;
-using namespace Eigen;
-using namespace grid_map;
+#include <vector>
+
+#include "grid_map_core/Polygon.hpp"
+
+// using namespace std;
```
maximilianwulf, line 20: As above.

maximilianwulf:
> @maximilianwulf Does it make sense to split this PR into small ones?

In general I would say yes, but most of the changes are only to the layout. As long as the unit tests pass I think we are on the safe side, @Marwan99?

grid_map_core/test/GridMapTest.cpp
```diff
@@ -6,24 +6,27 @@
...
 #include <math.h>
 
-using namespace std;
-using namespace grid_map;
+#include <vector>
+#include <string>
+
+#include "grid_map_core/GridMap.hpp"
+
+// using namespace std;
```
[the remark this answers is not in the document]
Marwan99: Just kept them as a reference while resolving namespaces, they are not needed now.

grid_map_core/src/Polygon.cpp
```diff
@@ -220,13 +225,15 @@ bool Polygon::offsetInward(const double margin)
   return true;
 }
 
-std::vector<Polygon> Polygon::triangulate(const TriangulationMethods& method) const
+std::vector<Polygon> Polygon::triangulate(const TriangulationMethods & method) const
 {
-  // TODO Add more triangulation methods.
+  // TODO(needs_assignment): Add more triangulation methods.
   // https://en.wikipedia.org/wiki/Polygon_triangulation
+  (void)(method);  // method parameter unused, should be removed once used.
```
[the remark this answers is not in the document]
Marwan99: Sure, will add it. A TODO needs to be assigned to someone (linter requirement), who should I put down? 

Also there are another 5 instances where parameters are not used and the same work around `(void)(param);` was used, I can list them if needed.  

On a similar note, in the existing TODOs I have added "someone" as a place holder, should that be updated as well?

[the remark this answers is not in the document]
maximilianwulf: Yes, add a placeholder.

SteveMacenski:
I'm good to merge when @Marwan99 CI passes (thanks for that!) and @maximilianwulf approves

Marwan99:
@SteveMacenski I need some help!
- Not sure why I am getting `readability/mixedcase` linting error since the whole message is upercase (last 2 failures). Any suggestions?
- Despite having `set(ament_cmake_copyright_FOUND TRUE)` the linter is still complaining about the lack of copyright messages. Do I have it in the wrong place? 

On another note the previous compiling error (before the last commit), was not happening locally on my machine and it was compiling perfectly fine, any suggestions why this might have happened?

SteveMacenski:
The readibility / mixed cases is in the cmake file, I think it just wants you to rename things on L4, 13. I think the legal one is from cpplint not the ament_copyright checker. Maybe there's a way to disable the specific `legal/copyright [5]` check in `cpplint`?

grid_map_core/CMakeLists.txt
```diff
@@ -81,50 +55,61 @@ target_link_libraries(${PROJECT_NAME}
...
-  test/EllipseIteratorTest.cpp
-  test/SubmapIteratorTest.cpp
-  test/PolygonIteratorTest.cpp
-  test/PolygonTest.cpp
-  test/EigenPluginsTest.cpp
-  test/SpiralIteratorTest.cpp
-  test/SlidingWindowIteratorTest.cpp
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  set(ament_cmake_copyright_FOUND TRUE)
```
naiveHobo, line 82:
```suggestion
  find_package(ament_lint_auto QUIET)
  if(ament_lint_auto_FOUND)
    # exclude copyright checks
    list(APPEND AMENT_LINT_AUTO_EXCLUDE
      ament_cmake_cpplint
      ament_cmake_copyright
    )
    ament_lint_auto_find_test_dependencies()

    # run cpplint without copyright filter
    find_package(ament_cmake_cpplint)
    ament_cpplint(
      FILTERS -legal/copyright
    )
  endif()
```

naiveHobo: This should disable all the copyright checks.

Marwan99: Thank you so much @naiveHobo! I found the FILTERS but I was struggling to get it working with ament_lint_auto.

grid_map_core/CMakeLists.txt
```diff
@@ -79,7 +79,21 @@ install(
...
   find_package(ament_lint_auto REQUIRED)
-  set(ament_cmake_copyright_FOUND TRUE)
+  find_package(ament_lint_auto QUIET)
+  if(ament_lint_auto_FOUND)
+    # exclude copyright checks
+    list(APPEND AMENT_LINT_AUTO_EXCLUDE
+      ament_cmake_cpplint
+      ament_cmake_copyright
+    )
+    ament_lint_auto_find_test_dependencies()
```
SteveMacenski, line 89: You call this 2x now

SteveMacenski:
Looks like 1 more linter error - also did your lines you added remove cpplint from core overall? I see one line where you exclude the copyright checks but it looks like you might also just turn it off

naiveHobo:
@SteveMacenski it removes copyright check (which is independent of cpplint) as well as cpplint (due to the separate `legal/copyright` checks) from `ament_lint_auto`. And cpplint is then run separately with the `legal/copyright' filter disabled.

SteveMacenski:
Ah ok, that makes sense. Can we make a note in the cmakelists to that effect so that's clear. I could see someone a year from now wanting to refactor that thinking we did something really silly

Marwan99:
If I'm understanding the error correctly, I can't think of a way to resolve the current cppcheck problem without breaking the API for the `float & GridMap::at` method. Is there a less destructive way of sorting this out?

SteveMacenski:
Please link to this function definition here

Marwan99:
https://github.com/Marwan99/grid_map/blob/db19acf8e2afb4ad01173415210f2fa07367974f/grid_map_core/src/GridMap.cpp#L232-L239

SteveMacenski:
mhm good question, that's a real issue. `at` returns a reference to the value in the map / unordered map, but then you access the reference and return some float value stored in the matrix. I googled around and it looks like the `operator()` in the `Eigen::MatrixXf` returns a `float &` so its also a reference to the class variable.

Therefore, I believe this is safe and can be given a `// NOLINT` (if that works for cppcheck as well) to be ignored. @maximilianwulf please verify that that makes sense.

naiveHobo:
Can this me merged quickly? I have ported `grid_map_costmap_2d` and `grid_map_octomap` to ros2 branching off this version. Would be great if we can close this asap.

SteveMacenski:
Need @maximilianwulf to give it another once over and approve

grid_map_core/src/GridMapMath.cpp
```diff
@@ -463,105 +552,122 @@ bool incrementIndexForSubmap(Index& submapIndex, Index& index, const Index& subm
...
+    return Index(
+      static_cast<int>(linearIndex) % bufferSize(
+        0), static_cast<int>(linearIndex) / bufferSize(0));
+  }
+  return Index(
+    static_cast<int>(linearIndex) / bufferSize(
+      1), static_cast<int>(linearIndex) % bufferSize(1));
+}
+
+// void getIndicesForRegion(
```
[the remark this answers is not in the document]
maximilianwulf: In general, we should just delete it. Can you delete it?

grid_map_core/CMakeLists.txt
maximilianwulf: @Marwan99 is there a reason for it?

maximilianwulf:
> Therefore, I believe this is safe and can be given a // NOLINT (if that works for cppcheck as well) to be ignored. @maximilianwulf please verify that that makes sense.

Did you manage to fix it without linting? I didn't find the corresponding change. 

Besides that, LGTM. Good job!

Marwan99:
@maximilianwulf This error was from cpp_check (which performs static code analysis, which is done with the linting checks), this was fixed by adding a comment telling the checker to ignore it, something similar to `// NOLINT`. So the API **was not** modified. You can find the fix in 507899f.

SteveMacenski:
just waiting on your OK @maximilianwulf

maximilianwulf:
Okay, got it. Thanks a lot.

SteveMacenski:
In@

PR #234, opened 2020-06-10
CI for ROS2

Marwan99:
CI for the ROS2 branch #219

Based on this [example](https://github.com/ros-perception/image_pipeline/blob/melodic/.circleci/config.yml ) as suggested by @SteveMacenski .

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2
+jobs:
+  debug_build_and_test:
+    docker:
+      - image: jwhitleywork/ros:foxy-ros-base-focal
+    steps:
+      - checkout
+      - run:
+          name: Set Up Container
+          command: |
+            apt update -qq && apt install -y build-essential cmake python3-colcon-common-extensions python3-rosdep libeigen3-dev
+            apt upgrade -y
+            source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
+            mkdir -p src/grid_map && mv `find -maxdepth 1 -not -name . -not -name src` src/grid_map/
+            rosdep update
+            rosdep install -y --ignore-src --from-paths src
+      - run:
+          name: Debug Build
+          command: |
+            source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
+            colcon build --parallel-workers 1
+      - run:
+          name: Run Tests
+          command: |
+            source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
+            colcon test --packages-select grid_map_core --parallel-workers 1
+            colcon test-result --verbose
+workflows:
+  version: 2
+  ros_build:
+    jobs:
+      - debug_build_and_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,3 +30,6 @@ workflows:
   ros_build:
     jobs:
       - debug_build_and_test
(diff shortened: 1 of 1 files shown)


maximilianwulf:
> Good first draft.
> 
> @maximilianwulf I don't have admin to setup circle for this org / repo. If you go to https://app.circleci.com/projects and give permission for Circle on the org, you can select this project and hit the enable button. After that, it should set things up for you and once we merge this, it will start working.

Okay, I will have to check internally. Will come back to you.

SteveMacenski:
OK, it doesn't auto-get permissions to the org or all the repos, you just select this one. It then has hooks to see when a PR is opened or merged to trigger builds, but it doesn't get access to overwrite content, I believe. Either way, you can review the matrix and you can see clearly that we just build and test, there's no writing to the repo. 

We need some kind of basic CI, but I'm not really picky one way or another. Circle is just what we use on navigation2 and image_pipeline, so I'm already having to deal with it daily.

remod:
CircleCI has access to the organization now. (I could not find an option to enable it for this repo only, which would be preferable)

SteveMacenski:
@remod https://app.circleci.com/projects/project-dashboard/github/ANYbotics you still need to click on "enable" on grid_maps I believe to get it setup. You'll know you're done when you go to settings -> webhooks of this repo and you see that there's a circle ci webhook auto generated.

remod:
Ok, thanks. :)

You are right, while setting up the "Hello World" example, the webhook has been created. Let me know if it does not work as desired.

SteveMacenski:
I see it set up now, thanks! 

@Marwan99 since we're only wanting to trigger this for ROS2 we need to add
```
          filters:
            branches:
              only: ros2
```

to the config so that we only trigger on ros2 branch and not for all builds if backporting. https://circleci.com/docs/2.0/workflows/ (control F for `branches` and you'll see the docs for it). Its unclear to me at this moment if that will only build the `ros2` branch, or will build anything PRing against the `ros2` branch. I think the best way to tell is to:
- Close this PR
- Add those lines to the config
- Create a new branch
- PR against `ros2` with the new branch.

Unfortunately, Circle is too smart for us to just toggle the PR to check

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+          command: |
+            source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
+            colcon test --parallel-workers 1
+            colcon test-result --verbose
+workflows:
+  version: 2
+  ros_build:
+    jobs:
+      - debug_build_and_test
+          filters:
```
SteveMacenski, line 33: This isn't the right place for the filters. it should be up in the job definition. See the docs for example.

```
workflows:
  version: 2
  commit:
    jobs:
      - test
      - deploy
  nightly:
    triggers:
      - schedule:
          cron: "0 0 * * *"
          filters:
            branches:
              only:
                - master
                - beta
    jobs:
      - coverage
```

Marwan99: Why wouldn't this work? As far as I understand this should build on every commit without a need for a schedule.

That's the example I followed: https://circleci.com/docs/2.0/configuration-reference/#filters-1

Please let me know if I am missing something.

SteveMacenski: I'm not saying schedule. I'm showing you where the filters yaml should go in this arbitrary example that included it.

PR #235, opened 2020-06-15
Ros2 CI PR re-submission

Marwan99:
Related to #234 #219

I did some testing on my and there were a couple main observations.

- Having branch filters with the `only` key prevented a PR from triggering circleci as it was on a branch which is not listed in the `only` list. So having a filter will prevent us from building and testing a PR before pushing to ros2 (unless we add the name of branch we are pushing from to the list, but that doesn't sound practical).
- Only a  push to a branch **with a circleci config** triggered the ci to start doing something.
- A branch with no config doesn't seem affected in anyway and PRs from branchs with no config look normal.

Given these observations I think it is safe to have no filters. And as far as I can tell other non-ros2 PRs and branches should not be affected by this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   debug_build_and_test:
     docker:
-      - image: jwhitleywork/ros:foxy-ros-base-focal
+      - image: ros:foxy
     steps:
       - checkout
       - run:
@@ -23,7 +23,7 @@ jobs:
           name: Run Tests
           command: |
             source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
-            colcon test --packages-select grid_map_core --parallel-workers 1
+            colcon test --parallel-workers 1
             colcon test-result --verbose
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2
+jobs:
+  debug_build_and_test:
+    docker:
+      - image: jwhitleywork/ros:foxy-ros-base-focal
+    steps:
+      - checkout
+      - run:
+          name: Set Up Container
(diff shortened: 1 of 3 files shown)


SteveMacenski:
@Marwan99 OK, I'm going to merge this. You should now submit a bogus PR (just add a space, whatever) to trigger CI on that build and then make changes required until it passes.

Marwan99:
That is strange circleci should have been triggered like in [this](https://github.com/Marwan99/grid_map/pull/11)

SteveMacenski:
I dont think it triggers when its not yet in the root branch.

SteveMacenski:
Also make sure the linters are on in auto test for all the packages

.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+      - checkout
+      - run:
+          name: Set Up Container
+          command: |
+            apt update -qq && apt install -y build-essential cmake python3-colcon-common-extensions python3-rosdep libeigen3-dev
+            apt upgrade -y
+            source `find /opt/ros -maxdepth 2 -name local_setup.bash | sort | head -1`
+            mkdir -p src/grid_map && mv `find -maxdepth 1 -not -name . -not -name src` src/grid_map/
+            rosdep update
+            rosdep install -y --ignore-src --from-paths src
```
Marwan99, line 16: Should we comment it out for now, till we start needing it, only specifying the packages needing it as we go, and at the end of the port return it to normal?

Marwan99: ^ Thats only for line 16

Marwan99:
```suggestion
```

SteveMacenski: No, we still need to install dependencies it requires since the docker container will be completely empty other than a basic ROS install. It needs to install deps to build

PR #237, opened 2020-06-23
[ROS2] grid_map_cv port

Marwan99:
related to  #219.

grid_map_cv package linted and ported.

This PR also includes a fix to `grid_map_core/CMakeLists.txt` which could be found in 7510c49. This fix is addressing the omission of an extra cmake config file and the omission of exporting the include directories.


Marwan99:
Unfortunately I have made the mistake of not creating this branch off `ANYbotics:ros2`, thats why there are too many commits in this PR, the relevant commits for this PR are the ones starting from 20/6 (last 6 commits). Please let me know if that is an issue.

Marwan99:
I think it is better to sort the branching issue first then resubmit this PR.

PR #240, opened 2020-06-24
Port grid_map_costmap_2d to ros2

naiveHobo:
Signed-off-by: Sarthak Mittal <sarthakmittal2608@gmail.com>


naiveHobo:
@SteveMacenski @Marwan99 need to add `nav2_costmap_2d` to pass Circle build but I'm guessing the grid_map ros2 branch is targeting foxy

Marwan99:
How about using the `rosplanning/navigation2` docker image as a temporary solution? Just until navigation2 gets added to foxy.

SteveMacenski:
I think the best answer is to add a `tools/ros2_dependencies.repos` file that contains the vcs address for map server so its built in the same workspace (isolate / remove all other packages from nav2 that aren't needed for map server)

grid_map_costmap_2d/include/grid_map_costmap_2d/grid_map_costmap_2d.hpp
```diff
@@ -6,7 +6,10 @@
  *	 Institute: ETH Zurich, ANYbotics
  */
 
-#pragma once
+#ifndef GRID_MAP_COSTMAP_2D__GRID_MAP_COSTMAP_2D_HPP_
+#define GRID_MAP_COSTMAP_2D__GRID_MAP_COSTMAP_2D_HPP_
 
 #include <grid_map_core/grid_map_core.hpp>
 #include <grid_map_costmap_2d/Costmap2DConverter.hpp>
```
SteveMacenski, line 13: Can we rename this file so that its camel case?

naiveHobo:
> I don't see anything to mention. Is there anything you want to highlight that you had to change by hand that might be worth specifically having me check?

I don't think anything needs to be looked over too much. There was a rostest which needed some work since I had to convert it to a normal gtest, it barely involved any launch stuff. Otherwise everything is pretty standard.

I need some help with the Circle CI config though, there are a couple of errors I can't figure out. Would be great if you could have a look.

SteveMacenski:
You didn't install the dependencies with rosdep for the underlay workspace.

SteveMacenski:
>  Definitely should work now :) 

All of my commits named that _never_ work now :-)

naiveHobo:
Haha, okay didn't work :laughing: I'm not sure where it's going wrong.

SteveMacenski:
try the -r flag

Marwan99:
Does the `UNDERLAY_WS` need to be sourced before running rosdep in the `OVERLAY_WS`?

SteveMacenski:
Yes, that is a good catch

naiveHobo:
@SteveMacenski @maximilianwulf ready to merge!

tools/ros2_dependencies.repos
```diff
@@ -0,0 +1,5 @@
+repositories:
```
maximilianwulf, line 1: @naiveHobo I know I am coming late to the party, but why did we add this dependency? Is it because of the costmaps? @na

maximilianwulf: Okay, I guess because of 

```
I think the best answer is to add a tools/ros2_dependencies.repos file that contains the vcs address for map server so its built in the same workspace (isolate / remove all other packages from nav2 that aren't needed for map server)
```

SteveMacenski: Yes, we need nav2_costmap_2d as a dependency but its not yet released since its super high in the dependency tree. Grid maps will at some point be also a dependency of navigation2 which creates a circular dependency so both nav2 and grid maps need to build certain elements of each other in CI to deal with a chicken and egg problem. Also lets both build master branch in case there are API breaking changes we can fix them more quickly

maximilianwulf: Okay, sounds good. Thanks.

PR #241, opened 2020-06-24
[ROS2] port grid_map_sdf

SteveMacenski:
- linted with uncrustify / cpplint
- changed long -> int64_t
- changing short -> int16_t
- adding inline to statics to suppress warnings for unused functions
- commented out pcl_ros function and headers because its not ported to ROS2 yet
- usual bits in cmake / package.xml
- rename h to hpp


SteveMacenski:
If this is OK, before merging I should open a ticket / comment in the ROS2 port that we need pcl_ros in ROS2 to finish this package

SteveMacenski:
@marwan99 please take a look

SteveMacenski:
Always fun when I see different things locally versus CI

grid_map_sdf/include/grid_map_sdf/distance_transform/pnmfile.h
```diff
@@ -18,20 +18,23 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307 USA
...
-#include "image.h"
-#include "misc.h"
+#include "grid_map_sdf/distance_transform/image.h"
+#include "grid_map_sdf/distance_transform/misc.h"
 
 namespace distance_transform
 {
 
-  const short BUF_SIZE = 256;
+  const int16 BUF_SIZE = 256;
```
Marwan99: I can see it compiling fine, but I am not sure if other compiler will be happy without "_t" in `int16`.

Marwan99:
> Always fun when I see different things locally versus CI

Gets more fun when things don't work locally but work in the CI :')

grid_map_sdf/include/grid_map_sdf/distance_transform/pnmfile.h
SteveMacenski: will do, sorry that one was missed in my second run through

SteveMacenski:
> Gets more fun when things don't work locally but work in the CI :')

That one's less usual to me haha

SteveMacenski:
Aight - GTG, @Marwan99 take a look

grid_map_sdf/src/SignedDistanceField.cpp
```diff
@@ -152,33 +167,33 @@ Vector3 SignedDistanceField::getDistanceGradientAt(const Position3& position) co
   j = std::max(j, 1);
   j = std::min(j, size_.y() - 2);
   k = std::max(k, 1);
-  k = std::min(k, (int)data_.size() - 2);
+  k = std::min(k, static_cast<int>(data_.size() - 2));
```
Marwan99, line 170:
```suggestion
  k = std::min(k, static_cast<int>(data_.size()) - 2);
```

PR #242, opened 2020-06-25
[ROS2] port grid_map_ros

Marwan99:
- rosbag replaced with rosbag2  (biggest change in this package port, see last 2 function definitions in GridMapRosConverter.cpp).
- Updated `grid_map_msgs::msg::GridMap` to the updated message definition.
- Updated `nav_msgs::OccupancyGrid` to `nav2_msgs::msg::Costmap`, consequently updated the API method function names.
- Linted and uncrustified.


Marwan99:
CI expected to fail initially due to dependency on `nav2_msgs` should be resolved once a fix is applied in #240 and the fix merged here.

Marwan99:
And CI is not showing up again... Am I doing something wrong?? This is branched off ros2.

grid_map_ros/package.xml
```diff
@@ -1,25 +1,37 @@
...
+  <buildtool_depend>ament_cmake</buildtool_depend>
+  <build_depend>grid_map_cmake_helpers</build_depend>
+
+  <depend>rclcpp</depend>
   <depend>grid_map_core</depend>
   <depend>grid_map_msgs</depend>
   <depend>grid_map_cv</depend>
   <depend>sensor_msgs</depend>
-  <depend>nav_msgs</depend>
+  <depend>nav2_msgs</depend>
```
SteveMacenski, line 21: You sure you don't also need nav_msgs?

SteveMacenski:
I think you really want to be using occupancy_grid rather than costmap message. The nav2_msgs/Costmap is really only used in 1 place for the costmap subscriber for the recovery server. The occupancy grid is the generic message published of the occupancy information and what rviz is displaying. 

I could see an argument for supporting both though

Marwan99:
I thought `nav_msgs` was replaced with `nav2_msgs`, at least comparing `OcuppancyGrid.msg` with `Costmap.msg` convinced me with this idea as they looked very similar from a brief look. I have reverted back to `OcuppancyGrid` and removed `Costmap`, I can bring `Costmap` back if both are needed, just let me know what you think.
 

>And CI is not showing up again... Am I doing something wrong?? This is branched off ros2. Could that be because I am running CI on my fork?
>
Removing circleci from my fork fixed the above ^.

SteveMacenski:
No, nav_msgs is still very much so used. nav2_msgs are an extension. 

I think supporting both would be fine - you already did the work for nav2_msgs/Costmap so may as well keep it, it could come in handy.

Marwan99:
I am struggling atm with getting rosbag2 to work, specifically unable to create a bag file. However I suspect there might be a bug in the rosbag2 API so I submitted an issue https://github.com/ros2/rosbag2/issues/448, but I am might be just doing something wrong. Anyhow, I should hopefully get some help from them on this issue soon.

Marwan99:
@SteveMacenski How long do you think we should wait for the rosbag2 issue to be fixed? Should we comment out the rosbag sections and move on for now?

SteveMacenski:
is there any movement on it? Where's the PR?

Marwan99:
CI is failing with `Unable to fetch some archives` when setting up the container. @SteveMacenski should we try a rerun?

SteveMacenski:
@Marwan99 I re-ran the job

Marwan99:
Sorry @SteveMacenski same error happening again. Am I doing something wrong?

SteveMacenski:
No, I think the server is just down for a little while. I looked and seems fine now, retriggered

SteveMacenski:
Cool! now running, just actual errors now :-)

Marwan99:
The `rcpputills` API has been updated and we need to use the updated version instead of the released version to make use of the new API function `rcpputils::fs::remove_all`. I have added `rcpputills` as a dependency and it seems to be building fine in the UNDERLAY_WS. However, when building grid_map_ros it seems that `rcpputills` is being included from the ROS_WS and not from the UNDERLAY_WS as excpected. The sourcing sequence LGTM and I can't figure out what is causing the problem. Any suggestions?

Since we will also need to add rosbag2 (as we now need the master branch for the bug fix) as a dependency and build it as part of the UNDERLAY_WS, why don't we just move to a ros2 rolling/nighlty image instead of foxy? I think it will make things simpler and potentially smoother in the long run.

SteveMacenski:
What is `rcpputills`? 

> However, when building grid_map_ros it seems that rcpputills is being included from the ROS_WS and not from the UNDERLAY_WS as excpected

Please be more specific here. What are you doing where and why. Are you saying that the foxy released versions aren't correct or the master ones have changed?

Marwan99:
I am referring to the [rcpputils](https://github.com/ros2/rcpputils) package.

For example on this [line](https://github.com/Marwan99/grid_map/blob/1c47169b2d9c7117d371c9848748f1795036c88e/grid_map_ros/test/GridMapRosTest.cpp#L82) in `saveLoad` test, I want to use the  [rcpputils::fs::remove_all](https://github.com/ros2/rcpputils/blob/5a1e09f1c9ae4efb4c726552ee8fa49740bf8f68/include/rcpputils/filesystem_helper.hpp#L549) function to remove any rosbags files before and after running the tests, this clean up must be done as a rosbag is being created in this test and the rosbag2 API will throw an error if a rosbag already exists with the same name, thus the need to perform pre and post test clean up.

[rcpputils::fs::remove_all](https://github.com/ros2/rcpputils/blob/5a1e09f1c9ae4efb4c726552ee8fa49740bf8f68/include/rcpputils/filesystem_helper.hpp#L549) (this function is different from [rcpputils::fs::remove](https://github.com/ros2/rcpputils/blob/5a1e09f1c9ae4efb4c726552ee8fa49740bf8f68/include/rcpputils/filesystem_helper.hpp#L523)) was recently added to both the master and foxy branches (~ 9 days ago) and is not part of the foxy released version AFAIK.

SteveMacenski:
huh, haven't run into that before, I thought you mean rclutils.

> rcpputils::fs::remove_all was recently added to both the master and foxy branches (~ 9 days ago) and is not part of the foxy released version AFAIK.

Thanks, that's a problem statement I can help with.  I see you added to the repos file, good, but it should be master branch if we're going to have to make things build from source to be up to date. For sourcing, source just the underlay, your issue is using the local setup bash files. Remove all local files and source the normal setup.bash so you get the full path of packages available. If you source /opt/ before building the underlay, then that's in the underlay path. If you then source the underlay, you'll get the opt stuff too if you do the full setup.bash. In general, I never use local_setup.bash for that reason. I want everything.

Marwan99:
Thanks for the help @SteveMacenski :)

Ready for review, to save you some time [grid_map_ros/src/GridMapRosConverter.cpp](https://github.com/ANYbotics/grid_map/pull/242/files#diff-04ba338d0b63a16343973dbcae4c8010) is the file with the most significant changes.

grid_map_ros/CMakeLists.txt
```diff
@@ -1,65 +1,48 @@
...
+cmake_minimum_required(VERSION 3.5)
 project(grid_map_ros)
 
-set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
-
 ## Find catkin macros and libraries
-find_package(catkin REQUIRED COMPONENTS
-  roscpp
+find_package(ament_cmake REQUIRED)
+find_package(grid_map_cmake_helpers REQUIRED)
```
SteveMacenski, line 6: I don't think you used the macros this exposes

.circleci/config.yml
```diff
@@ -25,20 +25,18 @@ jobs:
...
-            source $UNDERLAY_WS/install/local_setup.bash
+            colcon build --symlink-install --packages-up-to nav2_costmap_2d rcpputils rosbag2_cpp
+            source $UNDERLAY_WS/install/setup.bash
             cd $OVERLAY_WS && rosdep install -y --ignore-src --from-paths src
       - run:
           name: Debug Build
           command: |
-            source `find $ROS_WS -maxdepth 2 -name local_setup.bash | sort | head -1`
-            source $UNDERLAY_WS/install/local_setup.bash
+            source $UNDERLAY_WS/install/setup.bash
```
Marwan99, line 34: Is this line redundant? `underlay_ws` is already being sourced above. Same potential duplication with the sourcing in "Run Tests" below.

grid_map_ros/src/GridMapRosConverter.cpp
```diff
@@ -74,21 +95,26 @@ bool GridMapRosConverter::fromMessage(const grid_map_msgs::GridMap& message, gri
...
 
-void GridMapRosConverter::toMessage(const grid_map::GridMap& gridMap, const std::vector<std::string>& layers,
-                      grid_map_msgs::GridMap& message)
+void GridMapRosConverter::toMessage(
+  const grid_map::GridMap & gridMap, const std::vector<std::string> & layers,
+  grid_map_msgs::msg::GridMap & message)
 {
-  message.info.header.stamp.fromNSec(gridMap.getTimestamp());
-  message.info.header.frame_id = gridMap.getFrameId();
+  message.header.stamp = rclcpp::Time(gridMap.getTimestamp());
```
SteveMacenski, line 116: is `getTimestamp()` returning time in nanoseconds? The rclcpp::Time() input is nanoseconds (or the other API is seconds, nanoseconds)

SteveMacenski: Just overall do a once over and make sure the times are correct

grid_map_ros/CMakeLists.txt
Marwan99: done

grid_map_ros/CMakeLists.txt
```diff
@@ -1,65 +1,50 @@
...
+find_package(rclcpp REQUIRED)
```
maximilianwulf, line 7: I know they were not in alphabetical order, but could order these in alphabetical order?

grid_map_ros/CMakeLists.txt
```diff
@@ -1,65 +1,50 @@
...
   sensor_msgs
   nav_msgs
+  nav2_msgs
   std_msgs
   geometry_msgs
   cv_bridge
-  rosbag
-  tf
+  rosbag2_cpp
+  tf2
```
maximilianwulf, line 37: Nice.

grid_map_ros/package.xml
```diff
@@ -1,25 +1,38 @@
...
   <url type="website">http://github.com/anybotics/grid_map</url>
   <url type="bugtracker">http://github.com/anybotics/grid_map/issues</url>
   <author email="pfankhauser@anybotics.com">Péter Fankhauser</author>
-  <buildtool_depend>catkin</buildtool_depend>
-  <depend>roscpp</depend>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+  <build_depend>grid_map_cmake_helpers</build_depend>
+
+  <depend>rclcpp</depend>
```
maximilianwulf, line 16: Could you also order here alphabetically?

grid_map_ros/test/GridMapRosTest.cpp
```diff
@@ -72,14 +77,20 @@ TEST(RosbagHandling, saveLoad)
 
   EXPECT_FALSE(gridMapOut.exists(layer));
 
+  // Cleaning incase the previos bag was not removed
```
maximilianwulf, line 80:
```suggestion
  // Cleaning in case the previous bag was not removed
```

grid_map_ros/test/GridMapRosTest.cpp
```diff
@@ -99,9 +110,13 @@ TEST(RosbagHandling, saveLoadWithTime)
...
   EXPECT_FALSE(gridMapOut.exists(layer));
 
-  if (!ros::Time::isValid()) ros::Time::init();
-  // TODO Do other time than now.
-  gridMapIn.setTimestamp(ros::Time::now().toNSec());
+  // TODO(needs_assignment): Do other time than now.
+  rclcpp::Clock clock;
+  gridMapIn.setTimestamp(clock.now().nanoseconds());
+
+  // Cleaning incase the previos bag was not removed
```
maximilianwulf, line 117:
```suggestion
  // Cleaning in case the previous bag was not removed
```

grid_map_ros/test/GridMapRosTest.cpp
```diff
@@ -151,35 +169,133 @@ TEST(OccupancyGridConversion, roundTrip)
...
-      iterator != occupancyGrid.data.end(); ++iterator) {
+    iterator != occupancyGrid.data.end(); ++iterator)
+  {
     size_t i = std::distance(occupancyGrid.data.begin(), iterator);
-    EXPECT_EQ((int)*iterator, (int)occupancyGridResult.data[i]);
+    EXPECT_EQ(static_cast<int>(*iterator), static_cast<int>(occupancyGridResult.data[i]));
+  }
+}
+
+TEST(CostmapConversion, withMove)
```
maximilianwulf, line 218: Are these new unit tests?

grid_map_ros/src/GridMapRosConverter.cpp
maximilianwulf: I guess should be fine, it return is in nanoseconds: https://github.com/ANYbotics/grid_map/blob/02c5874deddfbf038241e6dc2438ef7b33f382ef/grid_map_core/include/grid_map_core/GridMap.hpp#L410

grid_map_ros/package.xml
SteveMacenski: Wow, even by my standards that's detail oriented :-)

grid_map_ros/test/GridMapRosTest.cpp
SteveMacenski: Yeah, he added both nav_msgs/occupancygrid and nav2_msgs/costmap conversions. We use nav2_msgs/costmap in a couple places to publish raw costmaps for other processes to use (primarily for recovery actions that need some view of the world but aren't in the same process as the local costmap / controller).

grid_map_ros/test/GridMapRosTest.cpp
Marwan99: Looks like the spell checker plugin is doing a fine job :')

grid_map_ros/CMakeLists.txt
Marwan99: Done!

grid_map_ros/package.xml
Marwan99: Done!

grid_map_ros/test/GridMapRosTest.cpp
maximilianwulf: That is very cool, thanks that you added them.

SteveMacenski:
@Marwan99 this OK to go in?

Marwan99:
Yeah lets merge!

PR #245, opened 2020-07-07
Merge Master into ROS2

remod:
Merging fixes from master into the ros2 branch.


SteveMacenski:
@remod where did you branch ros2 from? It looks like you have some merge conflicts and your branch was older than when we setup CI so CircleCI is failing.

maximilianwulf:
Hm, it only has the changes it should have. 

So CircleCi fails with: 

```
#!/bin/sh -eo pipefail
# No configuration was found in your project. Please refer to https://circleci.com/docs/2.0/ to get started with your configuration.
# 
# -------
# Warning: This configuration was auto-generated to show you the message above.
# Don't rerun this job. Rerunning will have no effect.
false
```

This looks to me more like a basic issue.

SteveMacenski:
That's why I'm saying I think its a base issue. There's even merge conflicts, I think their local version was really old. They should cherry pick or rebase.

remod:
@SteveMacenski `master` has got some fixes, and this PR just merges `master` into the `ros2` branch. As the `ros2` branch underwent a complete reformatting this leads to merge conflicts now and every time we want to update the branch with `master`. Is this the desired workflow? Does it make sense to reformat master as well?

SteveMacenski:
Perhaps, yes. I do think trying to cherry pick the commits over one by one would be fine.

remod:
Cherry-picking leads to the same merge conflicts, as you cannot apply commits 1-1 if the formatting changed...

SteveMacenski:
if you cherry-picked locally, you would have been asked to resolve these conflicts in the terminal. Since they appear here,  I don't think you did that correctly

remod:
Yes, it's clear that I can resolve the conflicts locally. However, it does not seem scale well if we have merge conflicts with every little fix we do on master. Does it make sense to reformat master as well?

SteveMacenski:
We could certainly do that, yes. The uncrustify errors can be automatically reformatted which will do the bulk of the work, but there's probably a solid day of linting for the cpplint compliance.

Keep in mind you'll still have merge conflicts regardless because the ROS APIs and structure have changed so much, but it would be lessor

Ryanf55:
I've marked this for 2.2.0 milestone, but needs an overhaul to only include the critical bugfixes. The rest can be kicked out to a later release.

PR #251, opened 2020-07-29
[ROS2] grid_map_loader port

Marwan99:
- Linted
- ROS related APIs updated

### DIFF
--- a/grid_map_loader/src/GridMapLoader.cpp
+++ b/grid_map_loader/src/GridMapLoader.cpp
@@ -11,7 +11,7 @@
 #include <grid_map_msgs/msg/grid_map.hpp>
 
 #include <chrono>
-
+#include <memory>
 #include <string>
 
 namespace grid_map_loader
@@ -63,7 +63,7 @@ void GridMapLoader::publish()
 {
   grid_map_msgs::msg::GridMap message;
   grid_map::GridMapRosConverter::toMessage(map_, message);
-  publisher_->publish(message);
+  publisher_->publish(std::make_unique<grid_map_msgs::msg::GridMap>(message));
   rclcpp::sleep_for(std::chrono::nanoseconds(duration_.nanoseconds()));
 }
 }  // namespace grid_map_loader
--- a/grid_map_loader/src/GridMapLoader.cpp
+++ b/grid_map_loader/src/GridMapLoader.cpp
@@ -24,11 +24,6 @@ GridMapLoader::GridMapLoader()
 
   rclcpp::QoS custom_qos(10);  // initialize to default
 
-  if (durationInSec > 0) {
-    // Expiration duration of persisted messasges.
-    custom_qos.lifespan(rclcpp::Duration::from_seconds(durationInSec));
-  }
-
   if (qos_transient_local_) {
     custom_qos.transient_local();  // Persist messages for “late-joining” subscriptions.
   }
--- a/grid_map_loader/src/GridMapLoader.cpp
+++ b/grid_map_loader/src/GridMapLoader.cpp
@@ -64,5 +64,6 @@ void GridMapLoader::publish()
   grid_map_msgs::msg::GridMap message;
   grid_map::GridMapRosConverter::toMessage(map_, message);
   publisher_->publish(message);
+  rclcpp::sleep_for(std::chrono::nanoseconds(duration_.nanoseconds()));
 }
 }  // namespace grid_map_loader
--- a/grid_map_loader/src/GridMapLoader.cpp
+++ b/grid_map_loader/src/GridMapLoader.cpp
@@ -7,22 +7,29 @@
  *
  */
 
-#include <grid_map_loader/GridMapLoader.hpp>
-#include <grid_map_msgs/GridMap.h>
+#include "grid_map_loader/GridMapLoader.hpp"
+#include <grid_map_msgs/msg/grid_map.hpp>
+
+#include <chrono>
 
 #include <string>
 
 namespace grid_map_loader
 {
 
-GridMapLoader::GridMapLoader(ros::NodeHandle nodeHandle)
-: nodeHandle_(nodeHandle)
+GridMapLoader::GridMapLoader()
+: Node("grid_map_loader")
 {
   readParameters();
-  publisher_ = nodeHandle_.advertise<grid_map_msgs::GridMap>(publishTopic_, 1, true);
+
+  rclcpp::QoS custom_qos(rclcpp::KeepLast(1));  // Buffer only last published message.
+  custom_qos.transient_local();  // Persist messages for “late-joining” subscriptions.
+  custom_qos.lifespan(duration_);  // Expiration duration of persisted messasges.
+
+  publisher_ = this->create_publisher<grid_map_msgs::msg::GridMap>(publishTopic_, custom_qos);
+
   if (!load()) {return;}
   publish();
-  ros::requestShutdown();
 }
 
 GridMapLoader::~GridMapLoader()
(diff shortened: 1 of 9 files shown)


Marwan99:
Build failing due to [grid_map_ros/include/grid_map_ros/message_traits.hpp](https://github.com/ANYbotics/grid_map/blob/ros2/grid_map_ros/include/grid_map_ros/message_traits.hpp) was missed in the grid_map_ros package port.

Working on fixing it.

SteveMacenski:
Type masquerading

grid_map_loader/src/GridMapLoader.cpp
```diff
@@ -7,21 +7,29 @@
...
 {
   readParameters();
-  publisher_ = nodeHandle_.advertise<grid_map_msgs::GridMap>(publishTopic_, 1, true);
-  if (!load()) return;
+
+  rclcpp::QoS custom_qos(rclcpp::KeepLast(1));  // Buffer only last published message.
+  custom_qos.transient_local();  // Persist messages for “late-joining” subscriptions.
+  custom_qos.lifespan(duration_);  // Expiration duration of persisted messasges.
+
+  publisher_ = this->create_publisher<grid_map_msgs::msg::GridMap>(publishTopic_, custom_qos);
```
SteveMacenski, line 29: I think the QoS should be configurable if they want one way or another. We do this in Nav2 / slam libraries too because some applications aren't compatible with this QoS

grid_map_loader/src/GridMapLoader.cpp
```diff
@@ -30,27 +38,31 @@ GridMapLoader::~GridMapLoader()
...
 }
 
 void GridMapLoader::publish()
 {
-  grid_map_msgs::GridMap message;
+  grid_map_msgs::msg::GridMap message;
   grid_map::GridMapRosConverter::toMessage(map_, message);
-  publisher_.publish(message);
-  duration_.sleep();
+  publisher_->publish(message);
```
SteveMacenski, line 66: Removed sleep? 

Also the publish should be a unique_ptr if possible. Can we change the `toMessage` signature?

grid_map_loader/src/GridMapLoader.cpp
```diff
@@ -64,5 +64,6 @@ void GridMapLoader::publish()
   grid_map_msgs::msg::GridMap message;
   grid_map::GridMapRosConverter::toMessage(map_, message);
   publisher_->publish(message);
+  rclcpp::sleep_for(std::chrono::nanoseconds(duration_.nanoseconds()));
```
SteveMacenski, line 67: or just duration_.sleep() ...

Marwan99: Not sure if this method is part of the [rclcpp::duration](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/duration.hpp) class.

SteveMacenski: That was unexpected, OK, makes sense then

grid_map_loader/src/GridMapLoader.cpp
Marwan99: By configurable you mean exposing them as parameters? The custom QOS implemented in nav2 that I managed to find were not configurable as far as I can tell (they were the ones in nav2_costmap_2d), can you please point me to an example?

Looking at the [QOS compatibilities tables](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-compatibilities) I don't see that the QOS config used here will cause any incompatibilities, as far as I can tell.

grid_map_loader/src/GridMapLoader.cpp
Marwan99: publish updated!

Changing `toMessage` to take in a unique_ptr as in `toMessage(..., unique_ptr<msg> message)`?

SteveMacenski: Uh that's not good `publisher_->publish(std::make_unique<grid_map_msgs::msg::GridMap>(message));` that's a big copy. you should change the signature of toMessage to return a unique pointer. For instance something really readable would be:

publisher_->publish(std::move(grid_map::GridMapRosConverter::toMessage(map_)));

grid_map_loader/src/GridMapLoader.cpp
SteveMacenski: https://github.com/ros-planning/navigation2/blob/master/nav2_costmap_2d/plugins/static_layer.cpp#L87 since this takes in the map which SLAM or localization could have different QoS settings. Its just helpful for the case you have a subscriber to these grid maps that you need to give to another process / node

grid_map_loader/src/GridMapLoader.cpp
Marwan99: Got it.

grid_map_loader/src/GridMapLoader.cpp
Marwan99: Should be configurable now. I have made the lifespan configurable only when a value bigger than 0  is passed to the `duration` param, If the user wants to keep the default QOS message lifespan, they will need to set the duration parameter to -1.

grid_map_loader/src/GridMapLoader.cpp
```diff
@@ -22,9 +22,16 @@ GridMapLoader::GridMapLoader()
...
   readParameters();
 
-  rclcpp::QoS custom_qos(rclcpp::KeepLast(1));  // Buffer only last published message.
-  custom_qos.transient_local();  // Persist messages for “late-joining” subscriptions.
-  custom_qos.lifespan(duration_);  // Expiration duration of persisted messasges.
+  rclcpp::QoS custom_qos(10);  // initialize to default
+
+  if (durationInSec > 0) {
+    // Expiration duration of persisted messasges.
+    custom_qos.lifespan(rclcpp::Duration::from_seconds(durationInSec));
```
SteveMacenski, line 29: I'm curious why you're setting lifespan, I don't typically see this QoS being messed with

grid_map_loader/src/GridMapLoader.cpp
```diff
@@ -61,9 +68,12 @@ bool GridMapLoader::load()
...
 void GridMapLoader::publish()
 {
-  grid_map_msgs::msg::GridMap message;
-  grid_map::GridMapRosConverter::toMessage(map_, message);
-  publisher_->publish(std::make_unique<grid_map_msgs::msg::GridMap>(message));
-  rclcpp::sleep_for(std::chrono::nanoseconds(duration_.nanoseconds()));
+  auto message = grid_map::GridMapRosConverter::toMessage(map_);
+  publisher_->publish(std::move(message));
+
+  if (durationInSec > 0) {
```
SteveMacenski, line 74: Not super clear on this - why is there a sleep here at all? Do we understand that?

grid_map_loader/src/GridMapLoader.cpp
Marwan99: In the original ros1 code:
-The publisher [latch flag](https://github.com/ANYbotics/grid_map/blob/07c8429b7bcabe182b9f2c203b2f10d3de583f3c/grid_map_loader/src/GridMapLoader.cpp#L21) is set to true.
-The [sleep](https://github.com/ANYbotics/grid_map/blob/c796c48d11fb36815a4366123089a2b080735292/grid_map_loader/src/GridMapLoader.cpp#L53) is used to keep the node running for a duration (instead of instant shutdown) to allow nodes that are created after the grid_map_loader node to still receive the published gridMap message (this was verified with @maximilianwulf on slack).

The lifespan is set to the `duration` parameter value along with setting the durability to `Transient local` to ensure that the message is available for the duration requested ` for late joining nodes.

grid_map_loader/src/GridMapLoader.cpp
Marwan99: See above for details. In short, keep this node running for late joining nodes to get the published message.

grid_map_loader/src/GridMapLoader.cpp
SteveMacenski: If you dont set the lifespan though the new nodes can get an arbitrarily old message as long as its the latest one.

SteveMacenski: Good or bad, not sure, it depends on the lifespan (he-he-he) of the grid map loader object and the context that its used in. For the nav2 map server for instance, I wouldn't want to set this but I'm not sure about this since I don't know off hand how its used.

Marwan99: Shall we open an issue or add it to the existing ros2 issue as something to test once we have grid_map_demo ported and we have a gridmap rosbag to play with?

SteveMacenski: The only use of this i see is here: https://github.com/ANYbotics/grid_map/blob/0f673c3600dae5ecf362d75e2e37e1617ff849c2/grid_map_loader/src/grid_map_loader_node.cpp

Should we be making the latched topic only temporary? It doesn't seem to me like that's the right option, this seems to act more like a map server than a time-out-valid temporary costmap

grid_map_ros/include/grid_map_ros/message_traits.hpp
```diff
@@ -1,73 +1,117 @@
...
   {
-    return m.info.header.stamp;
+    auto stamp = m.header.stamp;
+    return rclcpp::Time(stamp.sec, stamp.nanosec);
   }
 };
 
 }  // namespace message_traits
-}  // namespace ros
+}  // namespace grid_map
```
SteveMacenski, line 116: I think this namespace has to be `ros`

grid_map_loader/src/GridMapLoader.cpp
Marwan99: > Should we be making the latched topic only temporary?

What do you mean by temporary?

Are you suggesting changing this to be equivalent to map_server i.e. latching topic and always running and publishing to new subscribers?

SteveMacenski:
@Marwan99 see comment still above on the lifespan topic

grid_map_loader/src/GridMapLoader.cpp
Marwan99: @SteveMacenski Please see my question above, I think we both commented seconds away form each other :')

SteveMacenski: > What do you mean by temporary?

Setting lifespan sets a lifespan for the message so it becomes no longer valid after it expires, so its only latched for that duration. https://design.ros2.org/articles/qos_deadline_liveliness_lifespan.html. So I think you don't want to set lifespan at all and just have it be latched and the duration only be the sleeo.

Marwan99: Sure will remove it then. 

The [default lifespan](https://github.com/ros2/rmw/blob/163f0700d0df30b58c95a300d255f834a3805b8c/rmw/include/rmw/types.h#L430-L431) is set to 0, so I was just worried the message might expire spontaneously.

SteveMacenski: We dont set the lifespan at all in nav2 and I think the latching is working with the map server OK. We only make use of the basic 4 QoS where lifespan is part of the RT QoS

PR #253, opened 2020-08-17
[ros2] grid_map_visualization port

Marwan99:
Changes:
- Unsubscribing from topics is not available atm in rclcpp (according to [this](https://answers.ros.org/question/354792/rclcpp-how-to-unsubscribe-from-a-topic/)), so the subscription updating functionality in `GridMapVisualization` has been disabled for now.
- The parameters handled API in ROS2 is slightly different and so the parameters have been slightly modified, example:
  From this:

    ```
    grid_map_visualizations:
      - name: elevation_points
          type: point_cloud
          params:
            layer: elevation
      - name: elevation_grid
          type: occupancy_grid
          params:
            layer: elevation
            data_min: 0.08
            data_max: -0.16
    ```

    To this:
    ```
    grid_map_visualization:
        ros__parameters:
            grid_map_topic: /grid_map_simple_demo/grid_map
            grid_map_visualizations: [elevation_points, elevation_grid]
            elevation_points:
                type: point_cloud
                params:
                    layer: elevation
            elevation_grid:
                type: occupancy_grid
                params:
                    layer: elevation
                    data_min: 0.08
                    data_max: -0.16
    ```
  Is that okay??
- Since [rclcpp::publisher](http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Publisher.html) is a template, I have moved the publisher to live in the visualisation derived classes instead of having it in the `VisualizationBase` class.

### DIFF
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -78,8 +78,8 @@ private:
    */
   void updateSubscriptionCallback();
 
-  //! ROS nodehandle.
-  rclcpp::Node::SharedPtr nodeHandle_;
+  //! ROS node.
+  rclcpp::Node::SharedPtr node_;
 
   //! Parameter name of the visualizer configuration list.
   std::string visualizationsParameter_;
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -11,19 +11,19 @@
 #define GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
 
 #include <grid_map_msgs/msg/grid_map.hpp>
-#include <grid_map_visualization/visualizations/VisualizationFactory.hpp>
 #include <grid_map_visualization/visualizations/MapRegionVisualization.hpp>
+#include <grid_map_visualization/visualizations/OccupancyGridVisualization.hpp>
 #include <grid_map_visualization/visualizations/PointCloudVisualization.hpp>
 #include <grid_map_visualization/visualizations/VectorVisualization.hpp>
-#include <grid_map_visualization/visualizations/OccupancyGridVisualization.hpp>
+#include <grid_map_visualization/visualizations/VisualizationFactory.hpp>
 
 // ROS
 #include <rclcpp/rclcpp.hpp>
 
 // STD
+#include <memory>
 #include <string>
 #include <vector>
-#include <memory>
 
 namespace grid_map_visualization
 {
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -7,7 +7,8 @@
  *
  */
 
-#pragma once
+#ifndef GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
+#define GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
 
 #include <grid_map_msgs/GridMap.h>
 #include <grid_map_visualization/visualizations/VisualizationFactory.hpp>
@@ -20,23 +21,24 @@
 #include <ros/ros.h>
 
 // STD
+#include <string>
 #include <vector>
 #include <memory>
 
-namespace grid_map_visualization {
+namespace grid_map_visualization
+{
 
 /*!
  * Visualizes a grid map by publishing different topics that can be viewed in Rviz.
  */
 class GridMapVisualization
 {
- public:
-
+public:
   /*!
    * Constructor.
    * @param nodeHandle the ROS node handle.
    */
-  GridMapVisualization(ros::NodeHandle& nodeHandle, const std::string& parameterName);
+  GridMapVisualization(ros::NodeHandle & nodeHandle, const std::string & parameterName);
 
   /*!
    * Destructor.
@@ -47,10 +49,9 @@ class GridMapVisualization
    * Callback function for the grid map.
    * @param message the grid map message to be visualized.
    */
-  void callback(const grid_map_msgs::GridMap& message);
+  void callback(const grid_map_msgs::GridMap & message);
 
- private:
-
+private:
   /*!
    * Read parameters from ROS.
    * @return true if successful.
@@ -69,10 +70,10 @@ class GridMapVisualization
    * grid map to safe bandwidth.
    * @param timerEvent the timer event.
    */
-  void updateSubscriptionCallback(const ros::TimerEvent& timerEvent);
+  void updateSubscriptionCallback(const ros::TimerEvent & timerEvent);
 
   //! ROS nodehandle.
-  ros::NodeHandle& nodeHandle_;
+  ros::NodeHandle & nodeHandle_;
 
   //! Parameter name of the visualizer configuration list.
   std::string visualizationsParameter_;
@@ -99,4 +100,5 @@ class GridMapVisualization
   bool isSubscribed_;
 };
 
-} /* namespace */
+}  // namespace grid_map_visualization
+#endif  // GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -31,7 +31,7 @@ namespace grid_map_visualization
 /*!
  * Visualizes a grid map by publishing different topics that can be viewed in Rviz.
  */
-class GridMapVisualization
+class GridMapVisualization : public rclcpp::Node
 {
 public:
   /*!
@@ -50,12 +50,6 @@ public:
    * @param message the grid map message to be visualized.
    */
   void callback(const grid_map_msgs::msg::GridMap::SharedPtr message);
-
-  /*!
-   * Gets pointer to rclcpp::Node.
-   * @return shared pointer to rclcpp::Node object.
-   */
-  rclcpp::Node::SharedPtr get_node_ptr();
 
 private:
   /*!
@@ -77,9 +71,6 @@ private:
    * @param timerEvent the timer event.
    */
   void updateSubscriptionCallback();
-
-  //! ROS node.
-  rclcpp::Node::SharedPtr node_;
 
   //! Parameter name of the visualizer configuration list.
   std::string visualizationsParameter_;
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -10,7 +10,7 @@
 #ifndef GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
 #define GRID_MAP_VISUALIZATION__GRIDMAPVISUALIZATION_HPP_
 
-#include <grid_map_msgs/GridMap.h>
+#include <grid_map_msgs/msg/grid_map.hpp>
 #include <grid_map_visualization/visualizations/VisualizationFactory.hpp>
 #include <grid_map_visualization/visualizations/MapRegionVisualization.hpp>
 #include <grid_map_visualization/visualizations/PointCloudVisualization.hpp>
@@ -18,7 +18,7 @@
 #include <grid_map_visualization/visualizations/OccupancyGridVisualization.hpp>
 
 // ROS
-#include <ros/ros.h>
+#include <rclcpp/rclcpp.hpp>
 
 // STD
 #include <string>
@@ -36,9 +36,9 @@ class GridMapVisualization
 public:
   /*!
    * Constructor.
-   * @param nodeHandle the ROS node handle.
+   * @param parameterName The config parameter name.
    */
-  GridMapVisualization(ros::NodeHandle & nodeHandle, const std::string & parameterName);
+  explicit GridMapVisualization(const std::string & parameterName);
 
   /*!
    * Destructor.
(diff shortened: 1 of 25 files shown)


SteveMacenski:
> Unsubscribing from topics is not available atm in rclcpp 

You could reset the shared ptr to the subscriber to "unsubscribe" and re-create it to subscribe again.

The rest of the comments sounds good

Marwan99:
@SteveMacenski Please review.

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
+rclcpp::Node::SharedPtr
+GridMapVisualization::get_node_ptr()
+{
+  return nodeHandle_;
+}
+
 bool GridMapVisualization::readParameters()
 {
-  nodeHandle_.param("grid_map_topic", mapTopic_, string("/grid_map"));
+  nodeHandle_->declare_parameter("grid_map_topic", std::string("/grid_map"));
```
SteveMacenski, line 51: Please rename to `node`, nodehandles dont exist anymore and this won't age well

Marwan99: sure.

grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp
```diff
@@ -6,52 +6,61 @@
...
+      "did not find a 'height' parameter. Using default.",
+      name_.c_str());
   }
 
   return true;
 }
 
 bool FlatPointCloudVisualization::initialize()
 {
-  publisher_ = nodeHandle_.advertise<sensor_msgs::PointCloud2>(name_, 1, true);
```
Marwan99, line 40: Just noticed the publishers were latched, do we need them latched?

SteveMacenski: Good q- Max?

Marwan99:
@maximilianwulf please review.

maximilianwulf:
On it now.

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
-#include "grid_map_visualization/GridMapVisualization.hpp"
 #include <grid_map_core/GridMap.hpp>
 #include <grid_map_ros/GridMapRosConverter.hpp>
 
-using namespace std;
-using namespace ros;
+#include <string>
+#include <vector>
+#include <memory>
+#include <chrono>
```
maximilianwulf, line 15: Alphabetical order, can you check the other includes as well?

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
+      visualizationsParameter_.c_str(), type.c_str(), name.c_str());
   }
-
   return true;
 }
 
 bool GridMapVisualization::initialize()
 {
-  for (auto& visualization : visualizations_) {
+  for (auto visualization : visualizations_) {
```
maximilianwulf, line 126: Below you did not remove the `&`

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
+    node_->get_logger(),
+    "Grid map visualization received a map (timestamp %f) for visualization.",
+    rclcpp::Time(message->header.stamp).seconds());
   grid_map::GridMap map;
-  grid_map::GridMapRosConverter::fromMessage(message, map);
+  grid_map::GridMapRosConverter::fromMessage(*message, map);
 
-  for (auto& visualization : visualizations_) {
-    visualization->visualize(map);
+  for (std::size_t i = 0; i < visualizations_.size(); i++) {
```
maximilianwulf, line 170: Did it not work the other way around?

grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp
maximilianwulf: If possible yes, for visualization having it latched is nice.

grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp
```diff
@@ -6,52 +6,61 @@
...
+  publisher_ = node_->create_publisher<sensor_msgs::msg::PointCloud2>(name_, 1);
   return true;
 }
 
-bool FlatPointCloudVisualization::visualize(const grid_map::GridMap& map)
+bool FlatPointCloudVisualization::visualize(const grid_map::GridMap & map)
 {
-  if (!isActive()) return true;
```
maximilianwulf, line 46: I wonder if that should actually return false.

grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp
```diff
@@ -6,52 +6,61 @@
...
 }
 
 FlatPointCloudVisualization::~FlatPointCloudVisualization()
 {
 }
 
-bool FlatPointCloudVisualization::readParameters(XmlRpc::XmlRpcValue& config)
+bool FlatPointCloudVisualization::readParameters()
 {
-  VisualizationBase::readParameters(config);
```
maximilianwulf, line 28: Why don't we need this anymore here?

tools/ros2_dependencies.repos
```diff
@@ -2,7 +2,7 @@ repositories:
   ros-planning/navigation2:
     type: git
     url: https://github.com/ros-planning/navigation2.git
-    version: master
+    version: main
```
maximilianwulf, line 5: Interesting, do you know if this is this temporary or not?

maximilianwulf (commented):
Done. Some comments, but nonblocking ones. Thanks a lot again for the great effort!

Please ping me via mail if you addressed the comments.

tools/ros2_dependencies.repos
Marwan99: This should be permanent, I think other packages will be making this change as well AFAIK.

grid_map_visualization/include/grid_map_visualization/visualizations/VisualizationBase.hpp
```diff
@@ -50,67 +52,21 @@ class VisualizationBase
...
 
- protected:
-
-  /*!
-   * Get a visualization parameter as a string.
-   * @param[in] name the name of the parameter
-   * @param[out] value the string to set with the value.
-   * @return true if parameter was found, false otherwise.
-   */
-  bool getParam(const std::string& name, std::string& value);
```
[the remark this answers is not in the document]
Marwan99: Yes, getting parameters is handled differently in the [ROS2 API](http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html#a3158e5520c58531ae4d85a2e98cdf210).

grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp
Marwan99: Sure will change to false. For future reference, the logic is not going to be affected as the return of the `visualize` method is not checked (see below), but false makes more sense in this context.
https://github.com/ANYbotics/grid_map/blob/e3e00ff932bf8c6a30e1ed5a9e68cb0d5850d56c/grid_map_visualization/src/GridMapVisualization.cpp#L165-L167

grid_map_visualization/src/GridMapVisualization.cpp
Marwan99: Will reverted this change, removed while debugging and forgot to revert it back, my bad.

grid_map_visualization/src/GridMapVisualization.cpp
Marwan99: Will reverted this change, removed while debugging and forgot to revert it back, my bad.

grid_map_visualization/include/grid_map_visualization/visualizations/FlatPointCloudVisualization.hpp
```diff
@@ -6,29 +6,33 @@
...
-#pragma once
+#ifndef GRID_MAP_VISUALIZATION__VISUALIZATIONS__FLATPOINTCLOUDVISUALIZATION_HPP_
+#define GRID_MAP_VISUALIZATION__VISUALIZATIONS__FLATPOINTCLOUDVISUALIZATION_HPP_
 
-#include <grid_map_visualization/visualizations/VisualizationBase.hpp>
 #include <grid_map_core/GridMap.hpp>
 
 // ROS
-#include <ros/ros.h>
+// #include <rclcpp/rclcpp.hpp>
```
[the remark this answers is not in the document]
Marwan99: This is a bit trivial suggestion but should we keep just for including what we are using (linter usually asks for that for std libs but doesn't catch it for non-std libs).

grid_map_visualization/package.xml
```diff
@@ -1,21 +1,32 @@
...
+
   <depend>grid_map_core</depend>
-  <depend>grid_map_ros</depend>
   <depend>grid_map_msgs</depend>
-  <depend>tf</depend>
-  <depend>visualization_msgs</depend>
-  <depend>sensor_msgs</depend>
+  <depend>grid_map_ros</depend>
   <depend>nav_msgs</depend>
+  <depend>sensor_msgs</depend>
```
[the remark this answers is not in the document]
Marwan99: Done.

grid_map_visualization/src/GridMapVisualization.cpp
Marwan99: Done.

grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
```diff
@@ -47,10 +49,15 @@ class GridMapVisualization
...
    */
-  void callback(const grid_map_msgs::GridMap& message);
+  void callback(const grid_map_msgs::msg::GridMap::SharedPtr message);
 
- private:
+  /*!
+   * Gets pointer to rclcpp::Node.
+   * @return shared pointer to rclcpp::Node object.
+   */
+  rclcpp::Node::SharedPtr get_node_ptr();
```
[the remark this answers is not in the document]
Marwan99: I just needed to get a pointer to spin the node, not sure if its common practice though.
https://github.com/ANYbotics/grid_map/blob/c2a4dd3b52a17fc0c03d74e5bbe1117525774d82/grid_map_visualization/src/grid_map_visualization_node.cpp#L17-L19

maximilianwulf (approved):
Thanks, LGTM.

maximilianwulf:
Thanks, do you know why the pipeline is failing?

Marwan99:
rosbag2_cpp is failing to build, the error seems to be because the container is running out of memory is similar to what we were having before with grid_map_filters. I am testing fixes in a container locally atm.

Marwan99:
@SteveMacenski ready to merge.

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
-    : nodeHandle_(nodeHandle),
-      visualizationsParameter_(parameterName),
-      factory_(nodeHandle_),
-      isSubscribed_(false)
+GridMapVisualization::GridMapVisualization(const std::string & parameterName)
+: visualizationsParameter_(parameterName)
+  // isSubscribed_(false)
 {
-  ROS_INFO("Grid map visualization node started.");
+  node_ = std::make_shared<rclcpp::Node>("grid_map_visualization");
```
SteveMacenski, line 27: Why not make the grid map visualization class derive from node? That's pretty typical structure

grid_map_visualization/src/GridMapVisualization.cpp
```diff
@@ -6,165 +6,170 @@
...
+    std::bind(&GridMapVisualization::updateSubscriptionCallback, this));
   initialize();
 }
 
 GridMapVisualization::~GridMapVisualization()
 {
 }
 
+rclcpp::Node::SharedPtr
+GridMapVisualization::get_node_ptr()
```
SteveMacenski, line 44: Why not just let it access `node_`?

grid_map_visualization/src/grid_map_visualization_node.cpp
```diff
@@ -6,17 +6,17 @@
...
 
-  grid_map_visualization::GridMapVisualization gridMapVisualization(nodeHandle, "grid_map_visualizations");
-
-  ros::spin();
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  auto node = std::make_shared<grid_map_visualization::GridMapVisualization>(
+    "grid_map_visualizations");
+  rclcpp::spin(node->get_node_ptr());
```
SteveMacenski, line 19: If you derive from node, you can just spin `node->get_base_node_interface()`

SteveMacenski (changes requested):
Just some node cleanup and then looks good to merge

grid_map_visualization/src/GridMapVisualization.cpp
Marwan99: The visualisation classes (i.e [FlatPointCloudVisualization](https://github.com/Marwan99/grid_map/blob/visualization-port/grid_map_visualization/src/visualizations/FlatPointCloudVisualization.cpp) ..etc) needed access to the node pointer so they can declare/read parameters, log, check number of subscribers and create publishers. 

I was initially going down the route of making the grid map visualization class derive from node but I was unable to get a pointer to the node created to pass it on to the [visualisations classes](https://github.com/Marwan99/grid_map/tree/visualization-port/grid_map_visualization/src/visualizations), and the only solution I could find was to pass
 `rclcpp::node_interfaces::NodeLoggingInterface`, `rclcpp::node_interfaces::NodeGraphInterface`, `rclcpp::node_interfaces::NodeParametersInterface` and `rclcpp::node_interfaces::NodeBaseInterface` to each [visualization class](https://github.com/Marwan99/grid_map/tree/visualization-port/grid_map_visualization/src/visualizations) to be able to perform the things mentioned above. So instead I went with creating the node object inside the class to be be able to pass it around.

What do you think?

SteveMacenski: https://github.com/Marwan99/grid_map/blob/visualization-port/grid_map_visualization/include/grid_map_visualization/visualizations/MapRegionVisualization.hpp#L29 they derive from the base class, if that's a node, then each of the visualization is then itself a node.

Marwan99:
@SteveMacenski please have a look.

SteveMacenski (approved):

PR #255, opened 2020-08-25
Install .h header files

tomlankhorst:
.h header files missing in ROS Debian package @jelavice


maximilianwulf:
Thanks for fix, will import it.

PR #266, opened 2020-10-17
[ROS2] grid_map_demos port 3/3 

Marwan99:
Nodes/demos ported in this PR:

- resolution_change_demo 03fbae4
- interpolation_demo 96e9f18
- filters_demo port 37f1a3c

    Some cleanup was need for the grid_map_filters and grid_map_cv packages and the following changes were made:
        
    - Changes to their CMakeList.txt file (just minor things I missed while porting them... ops)  e140360
    - The design pattern used in the grid_map_filters parameter rely on being able to tell **whether a parameter has
    been added to the config file by the user or not**. However the default method used to read the parameters 
    [getParamImpl](https://github.com/ros/filters/blob/bfb591760164ecf71542778fdd59833afcca8aa7/include/filters/filter_base.hpp#L123-L146), which is called by [getParam](https://github.com/ros/filters/blob/bfb591760164ecf71542778fdd59833afcca8aa7/include/filters/filter_base.hpp#L160-L263) (this is the interface used to call `getParamImpl`) doesn't return whether a parameter has been found or not, also when a parameter is not found it assigns it a default value (that we can't change, e.g. "" for a string or 0 for an int).
    So I implemented a [new class](https://github.com/ANYbotics/grid_map/blob/e3a4a014de516603c7e88c17eb67d25550887c1d/grid_map_cv/include/grid_map_cv/utilities.hpp) to read parameters and behave like the parameter reading in ros1 filters as in [here](https://github.com/ros/filters/blob/54d4af832527e3e4729a28e4636dd4feacfb4069/include/filters/filter_base.h#L121-L136). Please let me know if there is a better way of handling this, also not sure if that's the best place for this file, what do you think? e3a4a01

- normal_filters_demo 805a990


SteveMacenski:
CI is still unhappy :(

SteveMacenski (commented):
Are all of the rviz config file change required? Many look like they don't really change much

Marwan99:
@SteveMacenski  Welcome back!
>CI is still unhappy :(

I think the CI is not happy atm due to this [issue](https://github.com/ros/filters/issues/42) in the filters. If you know someone we can ping for this issue to move things a bit that will be nice.

> Are all of the rviz config file change required? Many look like they don't really change much

You are correct, there might be some changes that have been saved when actually changing/fixing something, will look into them and revert any unnecessary ones.

SteveMacenski:
@maximilianwulf can you review?

maximilianwulf (approved):
Thank you. LGTM.

grid_map_core/src/GridMapMath.cpp
```diff
@@ -658,7 +658,8 @@ bool colorVectorToValue(const Eigen::Vector3i & colorVector, uint64_t & colorVal
 void colorVectorToValue(const Eigen::Vector3i & colorVector, float & colorValue)
 {
   uint64_t color = (colorVector(0) << 16) + (colorVector(1) << 8) + colorVector(2);
-  colorValue = static_cast<float>(color);
+  float * temp = reinterpret_cast<float *>(&color);
```
SteveMacenski, line 661: What's going on here?

grid_map_demos/src/normal_filter_comparison_node.cpp
```diff
@@ -74,32 +76,40 @@ void mapAverageFiltering(
 int main(int argc, char ** argv)
 {
   // Initialize node and publisher.
-  ros::init(argc, argv, "normal_filter_comparison_demo");
-  ros::NodeHandle nh("~");
-  ros::Publisher publisher = nh.advertise<grid_map_msgs::GridMap>("grid_map", 1, true);
+  rclcpp::init(argc, argv);
+  rclcpp::Node node("normal_filter_comparison_demo");
```
SteveMacenski, line 80: Can we make this a shared pointer? There have been times I've been burned in certain node operations not really working right because it expects a reference count to be incremented. ROS2 is really finnicky about shared pointers of certain objects. This might be completely fine, but its more defensive.

Marwan99: Updated the other demo nodes as well.

grid_map_core/src/GridMapMath.cpp
Marwan99: `colorValue = *reinterpret_cast<float*>(&color);` thats how it looks in [ROS1](https://github.com/ANYbotics/grid_map/blob/9cbb4eb315e3c4248849d0c1d9382928d0b4c3e6/grid_map_core/src/GridMapMath.cpp#L569). I have changed to static_cast during porting because the linter was complaining, but when testing with one of the demo nodes something didn't work, so reverted it back and managed to keep the linter happy and fixed the issue I was having.

SteveMacenski (approved):

PR #270, opened 2020-11-11
Fix CI for ROS2

mitsudome-r:
This fixes CI failure for ros2 branch.

Changes made:
* removed unnecessary dependencies in repos file (rcpputils, rosbag2, pcl_msgs can be installed by rosdep)
* fixed test failure in grid_map_ros package

### DIFF
--- a/grid_map_ros/test/GridMapRosTest.cpp
+++ b/grid_map_ros/test/GridMapRosTest.cpp
@@ -79,6 +79,7 @@ TEST(RosbagHandling, saveLoad)
   // Cleaning in case the previous bag was not removed
   rcpputils::fs::path dir(pathToBag);
   rcpputils::fs::remove_all(dir);
+  rcpputils::fs::create_directories(dir);
 
   EXPECT_TRUE(GridMapRosConverter::saveToBag(gridMapIn, pathToBag, topic));
   EXPECT_TRUE(GridMapRosConverter::loadFromBag(pathToBag, topic, gridMapOut));
@@ -116,6 +117,7 @@ TEST(RosbagHandling, saveLoadWithTime)
   // Cleaning in case the previous bag was not removed
   rcpputils::fs::path dir(pathToBag);
   rcpputils::fs::remove_all(dir);
+  rcpputils::fs::create_directories(dir);
 
   EXPECT_TRUE(GridMapRosConverter::saveToBag(gridMapIn, pathToBag, topic));
   EXPECT_TRUE(GridMapRosConverter::loadFromBag(pathToBag, topic, gridMapOut));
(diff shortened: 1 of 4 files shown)


SteveMacenski:
@marwan99 can you review?

grid_map_ros/test/GridMapRosTest.cpp
```diff
@@ -79,6 +79,7 @@ TEST(RosbagHandling, saveLoad)
   // Cleaning in case the previous bag was not removed
   rcpputils::fs::path dir(pathToBag);
   rcpputils::fs::remove_all(dir);
+  rcpputils::fs::create_directories(dir);
```
Marwan99, line 82: rosbag2_cpp already creates the required directory see [here](https://github.com/ros2/rosbag2/blob/90a67598c90b06a3ce02f92661f2dba38b24b51e/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L106), however this is not available in the binaries yet, that's why rosbag2_cpp is currently being built from source.

So I dont think this change is necessary and rosbag2_cpp should remain being built from source till the binaries are updated or the CI changed to use nightly builds instead of Foxy.

Marwan99 (changes requested):
The pybind11_vendor CI issue has already been addressed in #266 specifically in this [commit](https://github.com/ANYbotics/grid_map/pull/266/commits/6e3305664854b27946ea0921b022e327a6c4d0d5). I think we might need to keep rosbag2 building from source (see comment below), so you can potentially use the proposed fix in this PR to get the fix merged soon.

Otherwise LGTM :)

grid_map_ros/test/GridMapRosTest.cpp
mitsudome-r: I see.
Unless this is backported to foxy branch in rosbag2 repository, I don't think it would be available by binaries until the next ROS release.
Do you have any plans to do that by any chance?

mitsudome-r: If this is the only feature you need from the newest rosbag2 branch and if your ROS2 porting target is Foxy, I would suggest the following:
1. Create a PR to backport create_directory feature into foxy in rosbag2 repository and keep the code in grid_map.
1. I have a concern that they might reject the proposal to backport the commit because the change makes existing directory path unacceptable as input anymore, which might confuse the foxy users all over the world. In that case, I think we should add create_directories function in to the code in grid_map as in my MR until Galactic is release (in May 2021).

maximilianwulf:
So ready for merge?

grid_map_ros/test/GridMapRosTest.cpp
Marwan99: That's the only feature currently being used from the master branch, so your current proposal makes more sense to me.

Can you also please update the building instructions in the README?

mitsudome-r: Thanks!
I have updated the README

mitsudome-r:
I think this is ready to merge now, but maybe @Marwan99 can take a final look over the modified README and confirm.

Marwan99:
LGTM.

SteveMacenski (approved):

Issue #272, opened 2020-11-23
getIndex() Unable to Obtain Index for Maximum Negative Length of Map


BruceChanJianLe:
My grid_map is geometry size is as below:

```cpp
map_.setGeometry(grid_map::Length(1.0, 1.0), 0.05, grid_map::Position(0.0, 0.0));
```

When my position is -0.5 (return from getClosestPositionInMap()), which is still valid in the map, I get the error of 
```bash
/usr/include/eigen3/Eigen/src/Core/DenseCoeffsBase.h:365: Eigen::DenseCoeffsBase<Derived, 1>::Scalar& Eigen::DenseCoeffsBase<Derived, 1>::operator()(Eigen::Index, Eigen::Index) [with Derived = Eigen::Matrix<float, -1, -1>; Eigen::DenseCoeffsBase<Derived, 1>::Scalar = float; Eigen::Index = long int]: Assertion `row >= 0 && row < rows() && col >= 0 && col < cols()' failed.
```
However, if my position is `0.5` (return from getClosestPositionInMap()), I do not get the error, and I am able to obtain the index.

I specifically mentioned getClosestPositionInMap() function as I am not sure whether should the getClosestPositionInMap() return something like `-0.49` instead of `-0.5`, or is this a bug inside of getIndex().

Below code will fail when any of the x or y position is `-0.5` which is the maximum negative length of the map.
```cpp
if(map_.getIndex(points_["point_a"], index_a_)
&& map_.getIndex(points_["point_b"], index_b_))
 {
    // Loop through line region
    for(grid_map::LineIterator itr(map_, index_a_, index_b_); !itr.isPastEnd(); ++itr)
    {
        // Change value from 0.0 to 1.0
        map_.at("type", *itr) = 1.0;
        // Publish map
        publish();
    }
}
```

Thank you for taking your time to look into this matter:D

maximilianwulf:
Hey @BruceChanJianLe, thank you for submitting the bug. Could you submit a short unit test  into this file: https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/test/GridMapTest.cpp. I guess you can copy the code from above.

BruceChanJianLe:
Hi @maximilianwulf, thanks for getting back. I have submitted a short unit test as requested in https://github.com/ANYbotics/grid_map/pull/275. Please have a look at the failing test `NegativeClosestPositionInMap`. Thank you:D

maximilianwulf:
[...]

PR #275, opened 2020-12-02
update unit test

BruceChanJianLe:
Update unit test with PositiveClosestPositionInMap and NegativeClosestPositionInMap.

This is regarding issue #272 where I was requested to submit a unit test.

Hope this helps to discover the bug. Thank you.


maximilianwulf:
Thank you, the unit tests fail, right?

BruceChanJianLe:
> Thank you, the unit tests fail, right?

You're welcome! Indeed the unit test failed because of the bug where the return negative value (length of the grid map) is invalid when getIndex() function is called.

PR #293, opened 2021-04-07
Missing delete

manixq:
Added missing delete in getPlanarSignedDistanceField


maximilianwulf:
Thank you for the MR, we will include it.

Issue #283, opened 2021-01-14
ROS2 port remaining action items


Marwan99:
- [x] octomap_to_gridmap  demo node and launch files have been ported, however it has not been tested as `octomap_server`  has not been ported yet.
- [x] Finalise the `pcl_ros` related section in `grid_map_sdf` #243.
- [x] ~Once `pcl_ros` binaries are available, grid_map should be ready to release.~ no longer a dependency

SteveMacenski:
Renaming just to be a little more descriptive for passersby

maximilianwulf:
Thank you for the issue.

msy22:
How usable is the Foxy/gridmap port in it's current state? It appears that the `grid_map_sdf` and `grid_map_pcl` modules both depend on `pcl_ros`. But if neither of those are required, are all the other modules fully functional?

SteveMacenski:
Yes, that is my understanding as well! You can definitely use it in Foxy and if you run into any issues, let us know!

SteveMacenski:
There's a PR resolving that last item, so I think that means we can release binaries @Marwan99 @maximilianwulf ?

Marwan99:
Can binaries be released before the first item on the list is resolved?

> octomap_to_gridmap demo node and launch files have been ported, however it has not been tested as octomap_server has not been ported yet.

FYI, there has been some movement on the octomap_mapping [port](https://github.com/OctoMap/octomap_mapping/issues/76) lately.

SteveMacenski:
What's the scope of the octomap stuff? Do we actually need octomap headers for something or just launching in a launch file? Could that one be temporarily disabled until ported?

SteveMacenski:
To release to galactic (or any ROS2) we need the filters to be exporting its dependencies https://github.com/ros/filters/pull/50

Marwan99:
> What's the scope of the octomap 

I think this [launch file](https://github.com/ANYbotics/grid_map/blob/c5d985a5f7a9ce3a5a944a74164270f01ae24996/grid_map_demos/launch/octomap_to_gridmap_demo_launch.py#L35-L43) is the only thing that needs octomap_mapping, its already disabled there.

SteveMacenski:
OK, I think  filters is the only blocker then for attempting a release

maximilianwulf:
Hey, releasing binaries definitely sounds great. How does it work in the ROS2 world?

SteveMacenski:
[...]

Issue #243, opened 2020-06-25
[ROS2] Add back in pcl methods in grid_map_sdf once pcl_ros is available in ROS2


SteveMacenski:
currently pcl_ros is not available in ROS2 so the imported headers and conversion functions in grid_map_sdf cannot be fully ported at this time, but all the rest of it is.

Once pcl_ros is ported to ROS2, we can enable this.

maximilianwulf:
Is the ROS2 navigation group going to do it? 
Maybe we have to check if the package can work independent of `pcl_ros`.

maximilianwulf:
Related to #241 .

SteveMacenski:
I'm not sure off hand, one project at a time, pcl_ros isn't a small package. I also am a maintainer on that as well so I can help with getting things pushed through but we simply haven't discussed this in the context of the navigation working group

maximilianwulf:
Okay, thanks. I will check if we can have this package independent of `pcl_ros`.

Marwan99:
`pcl_ros` is now available in case you decide to continue using it.

I think this also needs to be added to the list in #283.

SteveMacenski:
pcl ros for ROS 2 is still not available - we haven't ported it yet

mitsudome-r:
@SteveMacenski 
Hi, I'm using this package for my project, and it also helps us if this is released as binary since it will reduce our build time. I was looking to see if there is anything I can help, and I've noticed that there could be alternative solution instead of waiting pcl_ros to be released. Could you take a look at https://github.com/ANYbotics/grid_map/pull/299, and see if the solution works for you?

SteveMacenski:
That works for me! Closing ticket since PR resolves

PR #299, opened 2021-05-13
[ros2] fix convertToPointCloud function to use PCL instead of pcl_ros

mitsudome-r:
In https://github.com/ANYbotics/grid_map/issues/283 and https://github.com/ANYbotics/grid_map/issues/243, grid_map_sdf package was blocked due to pcl_ros binary not being released.
However, since grid_map_sdf only uses pure PCL library API, the dependency can be replaced with libpcl-all-dev

This PR fixes convertToPointCloud function to depend on PCL instead of pcl_ros.

This way, grid_map_ros does not have to wait for pcl_ros to be released as binary package every time new ROS distribution is released.

### DIFF
--- a/grid_map_sdf/package.xml
+++ b/grid_map_sdf/package.xml
@@ -16,12 +16,12 @@
   <build_depend>grid_map_cmake_helpers</build_depend>
 
   <depend>grid_map_core</depend>
-  <!-- <depend>pcl_ros</depend> -->
+  <depend>libpcl-all-dev</depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_lint_auto</test_depend>
-  
+
   <export>
     <build_type>ament_cmake</build_type>
   </export>
--- a/grid_map_sdf/CMakeLists.txt
+++ b/grid_map_sdf/CMakeLists.txt
@@ -6,7 +6,7 @@ project(grid_map_sdf)
 find_package(ament_cmake REQUIRED)
 find_package(grid_map_cmake_helpers REQUIRED)
 find_package(grid_map_core REQUIRED)
-#find_package(pcl_ros REQUIRED)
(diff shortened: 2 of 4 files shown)


grid_map_sdf/package.xml
```diff
@@ -16,12 +16,12 @@
   <build_depend>grid_map_cmake_helpers</build_depend>
 
   <depend>grid_map_core</depend>
-  <!-- <depend>pcl_ros</depend> -->
+  <depend>libpcl-all-dev</depend>
```
SteveMacenski, line 19: Is there a specific submodule(s) of PCL this is using? It would be good to specify just the ones you need rather than installing ALL of PCL including things like visualizers and debuggers

SteveMacenski (commented):
I approve in concept, it just needs to use less of PCL's dependencies for smaller compute platforms

grid_map_sdf/package.xml
mitsudome-r: I'm not aware of a rosdep key available to install part of pcl. (we could still make it to `libpcl-all` instead of `dev` though)
https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L4093

This installs all pcl, but I think only required module (common module) is linked since it is specified with the following lines in the CMakeLists.txt.
`find_package(PCL REQUIRED COMPONENTS common)`

mitsudome-r: FYI, `pcl_ros` used to have [dependency to `libpcl-all-dev`](https://github.com/ros-perception/perception_pcl/blob/6900f7cf650e6c0df2aef45e0011833905b0ba9e/pcl_ros/package.xml#L38) in ROS1, so it's very likely to have the same dependency in ROS2. Therefore, I would say this PR isn't increasing the number of installed packages compared to ROS1 implementation.

SteveMacenski (approved):

SteveMacenski:
Thanks for the contribution!

PR #305, opened 2021-06-01
Update PR from master into melodic-devel

maximilianwulf:
Merging fixes from master into melodic-devel branch.


maximilianwulf:
@Jaeyoung-Lim fyi. I think these changes should not break melodic, in case they do please ping me.

Jaeyoung-Lim:
@maximilianwulf Thanks, probably this points to the problem that we need a ci for melodic.

I will add this using github actions

PR #306, opened 2021-06-15
Minor fix in getClosestPosition to avoid wrong indices

mmattamala:
Hi, this is a minor fix when using `getClosestPosition()` with positions outside the grid map.

The actual implementation used the `std::epsilon` value to avoid this but I still observed some issues when calling `getIndex()`. I used the same factor of 10 for epsilon used [here](https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/src/GridMapMath.cpp#L250) and seems to do the trick (I guess that just using epsilon is still too tight)

### DIFF
--- a/grid_map_core/src/GridMap.cpp
+++ b/grid_map_core/src/GridMap.cpp
@@ -719,18 +719,18 @@ Position GridMap::getClosestPositionInMap(const Position& position) const {
   const Position3 bottomRightCorner(position_.x() - halfLengthX, position_.y() - halfLengthY, 0.0);
 
   // Find constraints.
-  const double maxX = topRightCorner.x();
+  const double maxX = topLeftCorner.x();
   const double minX = bottomRightCorner.x();
-  const double maxY = bottomLeftCorner.y();
+  const double maxY = topLeftCorner.y();
   const double minY = bottomRightCorner.y();
 
   // Clip to box constraints and correct for indexing precision.
   // Points on the border can lead to invalid indices because the cells represent half open intervals, i.e. [...).
-  positionInMap.x() = std::fmin(positionInMap.x(), maxX - std::numeric_limits<double>::epsilon());
-  positionInMap.y() = std::fmin(positionInMap.y(), maxY - std::numeric_limits<double>::epsilon());
+  positionInMap.x() = std::fmin(positionInMap.x(), maxX - 10.0 * std::numeric_limits<double>::epsilon());
+  positionInMap.y() = std::fmin(positionInMap.y(), maxY - 10.0 * std::numeric_limits<double>::epsilon());
 
-  positionInMap.x() = std::fmax(positionInMap.x(), minX + std::numeric_limits<double>::epsilon());
-  positionInMap.y() = std::fmax(positionInMap.y(), minY + std::numeric_limits<double>::epsilon());
+  positionInMap.x() = std::fmax(positionInMap.x(), minX + 10.0 * std::numeric_limits<double>::epsilon());
+  positionInMap.y() = std::fmax(positionInMap.y(), minY + 10.0 * std::numeric_limits<double>::epsilon());
 
   return positionInMap;
 }


maximilianwulf:
Thank you very much. Could you check if you could add this scenario as a unit test as well? [This test](https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/test/GridMapTest.cpp#L158) should give some inspiration.

grid_map_core/src/GridMap.cpp
```diff
@@ -719,18 +719,18 @@ Position GridMap::getClosestPositionInMap(const Position& position) const {
   const Position3 bottomRightCorner(position_.x() - halfLengthX, position_.y() - halfLengthY, 0.0);
 
   // Find constraints.
-  const double maxX = topRightCorner.x();
+  const double maxX = topLeftCorner.x();
```
maximilianwulf, line 722: Both should be equal. Is this change for better readability?

mmattamala: Yes, indeed. I think it was clearer this way because it recalls the 2 corner points of the bounding box: _max_ values are the top-left ones while _min_ values the bottom-right's

mmattamala:
Thanks @maximilianwulf. I'll add the unit test and will ask your re-review.

mmattamala:
Hi @maximilianwulf, thanks for asking for the test, it made me realize the solution I was proposing is not solving the real problem.

I pushed [a test (`ClipToMap3`)](https://github.com/ANYbotics/grid_map/blob/9fa343f16df24003541d9c5b1333d1e7779de958/grid_map_core/test/GridMapTest.cpp#L355) that replicates the same settings that caused the original failure on my code. It is a 4x4m map with 0.02m resolution (200x200 cells).

It breaks when trying to clip a point to the lower bounds and clips the indices to 200 instead of 199 (I attached the output below). It doesn't break with the settings in `ClipToMap2` which uses a 1x1m, 0.05 map (20x20 cells).

I ran out of places to look at, but if you have any suggestions I can go further.

**Output:**
```sh
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:447: Failure
Expected equality of these values:
  expectedIndex.y()
    Which is: 199
  insideIndex.y()
    Which is: 200
closestInsidePosition 1.9999999999999998
-1.9999999999999998
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:450: Failure
Value of: isInside
  Actual: false
Expected: true
position is: 
 1.9999999999999998
-1.9999999999999998
 index is: 
  0
200

/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:493: Failure
Expected equality of these values:
  expectedIndex.y()
    Which is: 199
  insideIndex.y()
    Which is: 200
closestInsidePosition                  0
-1.9999999999999998
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:496: Failure
Value of: isInside
  Actual: false
Expected: true
position is: 
                  0
-1.9999999999999998
 index is: 
100
200

/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:516: Failure
Expected equality of these values:
  expectedIndex.x()
    Which is: 199
  insideIndex.x()
    Which is: 200
closestInsidePosition-1.9999999999999998
 1.9999999999999998
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:520: Failure
Value of: isInside
  Actual: false
Expected: true
position is: 
-1.9999999999999998
 1.9999999999999998
 index is: 
200
  0

/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:539: Failure
Expected equality of these values:
  expectedIndex.x()
    Which is: 199
  insideIndex.x()
    Which is: 200
closestInsidePosition-1.9999999999999998
                  0
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:543: Failure
Value of: isInside
  Actual: false
Expected: true
position is: 
-1.9999999999999998
                  0
 index is: 
200
100

/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:563: Failure
Expected equality of these values:
  expectedIndex.x()
    Which is: 199
  insideIndex.x()
    Which is: 200
closestInsidePosition-1.9999999999999998
-1.9999999999999998
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:564: Failure
Expected equality of these values:
  expectedIndex.y()
    Which is: 199
  insideIndex.y()
    Which is: 200
closestInsidePosition-1.9999999999999998
-1.9999999999999998
/home/matias/catkin_ws/src/grid_map/grid_map_core/test/GridMapTest.cpp:567: Failure
Value of: isInside
  Actual: false
Expected: true
position is: 
-1.9999999999999998
-1.9999999999999998
 index is: 
200
200
```

maximilianwulf:
Thank you, I will import it.

guichristmann:
Resuscitating what seems to be an old issue. I am running into a similar problem. It seems that `getClosestPositionInMap` returns positions that trigger an out of bounds exception. Applying the changes proposed in this PR and compiling grid_map from source seems to fix the problem.

I'm surprised more people don't run into the issue often? Unless I have something else quirky going on in my end.

PR #308, opened 2021-06-23
Replaced pcl_ros dependency with PCL for grid_map_pcl (branch: foxy-devel)

MattSYoung:
* Replaced pcl_ros dependency with PCL
* Updated changelog

### DIFF
--- a/grid_map_pcl/package.xml
+++ b/grid_map_pcl/package.xml
@@ -18,7 +18,6 @@
   <depend>grid_map_core</depend>
   <depend>grid_map_msgs</depend>
   <depend>grid_map_ros</depend>
-  <depend>pcl_ros</depend>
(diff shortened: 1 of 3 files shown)


grid_map_pcl/package.xml
```diff
@@ -18,7 +18,6 @@
   <depend>grid_map_core</depend>
   <depend>grid_map_msgs</depend>
   <depend>grid_map_ros</depend>
-  <depend>pcl_ros</depend>
   <depend>rclcpp</depend>
```
SteveMacenski, line 21: You must then replace the rosdep keys to include PCL

SteveMacenski (changes requested):

SteveMacenski (approved):

PR #309, opened 2021-06-23
Replaced pcl_ros dependency with PCL for grid_map_pcl (branch: ros2)

MattSYoung:
* Replaced pcl_ros dependency with PCL
* Updated changelog

### DIFF
--- a/grid_map_pcl/package.xml
+++ b/grid_map_pcl/package.xml
@@ -18,7 +18,6 @@
   <depend>grid_map_core</depend>
   <depend>grid_map_msgs</depend>
   <depend>grid_map_ros</depend>
-  <depend>pcl_ros</depend>
   <depend>rclcpp</depend>
   <depend>rcutils</depend>
   <depend>yaml-cpp</depend>
--- a/grid_map_pcl/package.xml
+++ b/grid_map_pcl/package.xml
@@ -21,6 +21,7 @@
   <depend>rclcpp</depend>
   <depend>rcutils</depend>
   <depend>yaml-cpp</depend>
+  <depend>libpcl-all</depend>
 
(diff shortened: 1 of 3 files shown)


grid_map_pcl/package.xml
```diff
@@ -18,7 +18,6 @@
   <depend>grid_map_core</depend>
   <depend>grid_map_msgs</depend>
   <depend>grid_map_ros</depend>
-  <depend>pcl_ros</depend>
```
SteveMacenski, line 21: You must then replace the rosdep keys to include PCL

SteveMacenski (changes requested):

grid_map_pcl/package.xml
MattSYoung: Forgive my ignorance, but which files should I edit to change the rosdep keys?

SteveMacenski: in this package.xml file, you have to add depends so that when you use rosdep it can properly install all of the dependencies. It is also requires so the build farm will work properly. https://github.com/ros/rosdistro see keys here. Look at the docs / tutorials on using rosdep / rosdistro.

MattSYoung: I've updated the package.xml file to have the correct dependency on `libpcl-all`. When it comes to the rosdep / rosdistro repo, I don't see any rosdep keys for anything grid_map related in `base.yaml`... should there already be a key there?

SteveMacenski (approved):

frapit:
I think the `libpcl-all` rosdep introduced in this PR is not sufficient, we need `libpcl-all-dev`. On a fresh `ros:foxy` container, `grid_map_pcl` won't compile otherwise. Same goes for the `ros2` branch.

PR #312, opened 2021-06-30
Fix mismatching version in sub-modules

wallner90:
Mismatching versions in sub-modules (1.6.2 vs. 2.0.0), may lead to issues with git-bloom-release.


devrite:
Fixes #314 for `foxy-devel`

Ryanf55:
This is no longer valid. `rolling` has version `2.1.0` on all packages today. https://github.com/ANYbotics/grid_map/blob/886ba24724052cbb55e616ee36651a7fa07e9714/grid_map_costmap_2d/package.xml#L5

And, `humble` has `2.0.0` on all packages today.
https://github.com/ANYbotics/grid_map/blob/d97bf39830442fe0556b2632eb213879d03a1c91/grid_map_costmap_2d/package.xml#L5

Closing this because `foxy` is EOL.

PR #318, opened 2021-08-12
Fix for clang build

kenji-miyake:
As I found this repository can't be built with `clang++`, I fixed errors.

### How to reproduce the problem.

```sh
git clone git@github.com:ANYbotics/grid_map.git -b ros2
cd grid_map
CXX=clang++ colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release
```

PR #331, opened 2021-12-10
Update namespace of to_yaml

wep21:


wep21:
This will be fixed by https://github.com/ANYbotics/grid_map/pull/318.

PR #350, opened 2022-05-24
Use default C++ and boost::placeholders::_N instead of deprecated _N

lucasw:
The default C++ will be 14 in noetic Ubuntu 20.04, 17 or later on newer systems- 17 is necessary to avoid issues with log4cxx 0.12.


maximilianwulf:
Thank you, should be fine. Processing it with the team.

grid_map_rviz_plugin/include/grid_map_rviz_plugin/modified/frame_manager.h
```diff
@@ -192,9 +192,11 @@ class FrameManager : public QObject
   template <class M>
   void registerFilterForTransformStatusCheck(tf2_ros::MessageFilter<M>* filter, Display* display)
   {
-    filter->registerCallback(boost::bind(&FrameManager::messageCallback<M>, this, _1, display));
+    filter->registerCallback(boost::bind(&FrameManager::messageCallback<M>, this,
```
maximilianwulf, line 195: With c++14 it is recommended to use lambdas, not?

grid_map_cv/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.5.1)
 project(grid_map_cv)
 
-set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
 add_compile_options(-Wall -Wextra -Wpedantic)
```
maximilianwulf, line 4: set(CMAKE_CXX_STANDARD 17)

?

lucasw: The idea here is that I assume every system this is building on now is >= 11 so on Ubuntu 20.04 the default will be 14, and on 22.04 it will be 17, there's no reason to hold it back at 11.

grid_map_rviz_plugin/include/grid_map_rviz_plugin/modified/frame_manager.h
lucasw: This seems like a smaller change for right now, would prefer to get this in and look into other options later, and I think still is backwards compatible back to c++11 if that is a concern for any other users (but possibly lambdas are too?  Would have to look into it).

PR #365, opened 2022-07-12
Do not iterate over the map when size is 0

Smiffe:
Hey,

I have noticed small bug within GridMapIterator. There are two situations in which GridMap object is empty (size is 0,0):
- When geometry of GridMap is not set, so it's by default (0,0)
- Due to very small length values, so size may be rounded to (0,0) https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/src/GridMap.cpp#L51

When using GridMapIterator as shown in README.md and demos https://github.com/ANYbotics/grid_map/blob/master/grid_map_demos/src/IteratorsDemo.cpp#L58, first iteration is always valid as `isPastEnd_` variable is `false` regardless of the GridMap's size https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/src/iterators/GridMapIterator.cpp#L20. Accessing empty GridMap leads to exception which crashes whole application as currently it's not catched or even documented.

I have added validation in GridMapIterator constructor which ensures GridMap is accessible. I have also added simple test to check if `isPastEnd_` variable initializes with `true` value when passed GridMap is empty.

### DIFF
--- a/grid_map_core/src/iterators/GridMapIterator.cpp
+++ b/grid_map_core/src/iterators/GridMapIterator.cpp
@@ -17,7 +17,7 @@ GridMapIterator::GridMapIterator(const grid_map::GridMap& gridMap)
   startIndex_ = gridMap.getStartIndex();
   linearSize_ = size_.prod();
   linearIndex_ = 0;
-  isPastEnd_ = false;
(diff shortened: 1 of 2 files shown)


maximilianwulf:
Thank you, will import it.

PR #383, opened 2023-01-18
Warning hunt

MaximilienNaveau:
Dear all,

I noticed many warning related to shadow variable (re-declaration of variable with the same name in 1 scope).
This can really cause trouble during the run so I fixed all of them.

One more reason to do that is that on my robot these warnings are necessarily treated as errors. It's a safety measure used to prevent bugs from happening.

One detail on this PR I reverted the include name from [filters (the ros package)](http://wiki.ros.org/filters) from ".hpp" to ".h". As the headers are still maintained in the ".h" form in ROS1 noetic.
This allows me to use your code in melodic while not perturbing the usage in noetic.


MaximilienNaveau:
@mgaertneratanybotics Would you mind reviewing this PR?
I think it's gonna be very quick

PR #384, opened 2023-01-23
fixed RViz crash if flat height layer is invalid

### LINKED ISSUE #161: rviz display crash when height layer is not available
`GridMap` throws exceptions when lookup of things fail but these are not caught in the `rviz` display which in return crashes `rviz` altogether.

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  GridMap::at(...) : No map layer 'elevation' available.
```

It happens to me when I'm changing HeightTransformer type with a layer name that's not available. Anyone found this before?

simongiegerich:
fixes #161 

Fixes these bugs:
-  RViz crashes if the Height Transformer is set to Flat while the Height Layer is invalid.
-  RViz crashes if the Color Transformer is set to None while the Color Layer is invalid.


fvina-anybotics (approved):
LGTM, thanks!

PR #394, opened 2023-06-08
 fix: initialize outputColor in ColorBlendingFilter.cpp

xaru8145:


Ryanf55:
This is already merged in `rolling` and `humble`. No need to merge.

Issue #371, opened 2022-09-03
Unable to build grid_map because can't find pcl_ros. Under ROS2 Foxy and using the branch ros2


pgarcia-dev:
Hello,

I post here the url where I explain the problem I'm facing and I wrote in ROS answers: https://answers.ros.org/question/405833/unable-to-build-grid_map-because-cant-find-pcl_ros/

Thanks

RaamRosik:
Use the branch foxy-devel

pgarcia-dev:
> Use the branch foxy-devel

Thank you! Now I was able to build grid_map And execute the demos 👌

alexMarFar:
Hi! I have the same problem but I cannot seem to solve it. 

I have tried following the Readme from ros2 branch and got the "Package 'pcl_ros' specified with --packages-up-to was not found" message. After finding this issue I tried with th foxy-devel branch, following the read me commands and even modifying the line "wget https://raw.githubusercontent.com/ANYbotics/grid_map/ros2/tools/ros2_dependencies.repos" to "wget https://raw.githubusercontent.com/ANYbotics/grid_map/foxy-devel/tools/ros2_dependencies.repos" in case there was a typo bu still get the "Package 'pcl_ros' specified with --packages-up-to was not found" message. 

Is there something I am doing wrong?

pgarcia-dev:
Hello Alejandra,
I finally solved the problem using the branch foxy-devel instead of the ros2 one.

I closed a ticket I opened on Ros answers: https://answers.ros.org/question/405833/unable-to-build-grid_map-because-cant-find-pcl_ros/

Best regards,
Pablo

PR #407, opened 2023-11-07
Update ros2_dependencies.repos

Claudio-Chies:
Fixes an which comes apprears during ````colcon build --symlink-install --packages-up-to pcl_ros``` which thre me ```Package 'pcl_ros' specified with --packages-up-to was not found``` (same issue as in [old Issue](https://github.com/ANYbotics/grid_map/issues/371)

### DIFF
--- a/tools/ros2_dependencies.repos
+++ b/tools/ros2_dependencies.repos
@@ -1,5 +1,5 @@
 repositories:
-  ros-planning/navigation2:
+  ros-perception/perception_pcl:
     type: git
-    url: https://github.com/ros-planning/navigation2.git
-    version: main
+    url: https://github.com/ros-perception/perception_pcl.git
+    version: ros2


Ryanf55:
I talked to Steve from NAV2, and for rolling, it will need to be compiled from source, and rosdep should use the `-r` flag because of circular dependencies in slam_toolbox. 

For humble and iron, we should use binaries for nav2 and not compile dependencies from source.

Ryanf55:
`perception_pcl` is available in the ROS index, and should be brought in through rosdep.
https://index.ros.org/p/perception_pcl/github-ros-perception-perception_pcl/#humble

Can you try that instead?

Ryanf55:
Humble CI is passing without compiling any dependencies from source. Because I got no response, and need to more forward with the release, I'm closing this.

PR #408, opened 2023-11-07
Update README.md branches for humble

Claudio-Chies:
Fixed README so it has the correct ros version and branches


Ryanf55:
Similar to https://github.com/ANYbotics/grid_map/pull/405

Ryanf55 (approved):
Thanks! We should do this for `iron` and `rolling` too! Would you either be able to submit a respective PR for `iron` and `rolling`, or review a PR if I do the work?

PR #418, opened 2024-02-03
Add dockerfile for rolling

Ryanf55:
# Purpose

Add a dockerfile, which makes it easier to build this library. Soon, ROS 2 will start updating to run on newer versions of Ubuntu than Ubuntu 22, which makes it hard for people who choose to stay on  22.04 and `humble`. 

# Risks

This may not build all the time due to ABI breaks, and the ROS docker images for rolling are not ABI stable. 

# Instructions

`docker build .`

# Future improvements

Get CircleCI to build the dockerfile instead of all the crazy things it's currently doing.


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+COPY . src/grid_map
+RUN ls src/grid_map
+
+SHELL ["/bin/bash", "-c"]
+RUN apt-get update \
+    && rosdep update
+
+RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
+    && rosdep install -y --ignore-src --from-paths src --skip-keys slam_toolbox
```
Ryanf55, line 23: @stevemacenski This is a proposed workaround in `grid_map` to deal with the fact that not all the keys resolve in `nav2` when we build it from source in `rolling`.
Do you have any way in NAV2 to ensure that there won't be new keys added in `rolling` that don't resolve? If that happens, it would break CI here.

If I use the `-r` flag globally, then I let contributors to grid_map contribute mis-typed keys. This is a real problem and happened in ArduPilot before: 
https://github.com/ArduPilot/ardupilot_ros/pull/23

SteveMacenski: Are you only building nav2 for costmap2d? If so, you could give rosdep only that package path to resolve and build only up to that.

Ryanf55: Both `costmap2d` and `nav2_msgs` are used in `grid_map`. Sounds like the risk of breakage to this line of code is very low.

.dockerignore
```diff
@@ -0,0 +1,5 @@
+.circleci/
+.github/
+.gitignore/
+LICENSE/
+README.md
```
wep21, line 5:
```suggestion
README.md

```

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+# Copy source code
+COPY . src/grid_map
+
+# Install dependencies
+SHELL ["/bin/bash", "-c"]
+RUN apt-get update && \
+    rosdep update && \
+    source /opt/ros/${ROS_DISTRO}/setup.bash && \
+    rosdep install -y --ignore-src --from-paths src --skip-keys slam_toolbox && \
```
wep21, line 30:
```suggestion
    rosdep install -y --ignore-src --from-paths src --skip-keys slam_toolbox --skip-keys gazebo_ros_pkgs --skip-keys turtlebot3_gazebo && \
```

Ryanf55: See my upstream PR for a longer term solution to ignore gazebo classic:
https://github.com/ros-planning/navigation2/pull/4095

wep21 (approved):
LGTM

PR #461, opened 2024-06-11
Fix the Indices of the inpaint filter

ipa-dmo:
The inpaint filter does not use the correct indices when creating the new grid map layer.

This issue is also included in the iron release.


Ryanf55:
Hello. Can you please target this to rolling. Then, we can backport it. Thanks for the submission.

Ryanf55:
Also, do you have any recommendations on how to write an automated test to verify the fix?

Ryanf55:
Even if not automated, do you have any recommendations on how I can verify this is correct, or reproduce the problem before this patch is applied? I am not yet familiar with this part of the codebase. 

Thanks!

PR #465, opened 2024-07-25
ci: update rolling ci

wep21:
- update build actions for ros2 rolling


Ryanf55 (approved):
Awesome, thanks for this!

PR #466, opened 2024-07-28
Add ROS 2 Jazzy backport mergify rule

Ryanf55:
# Purpose

Add  a mergify action such that we can backport to Jazzy


wep21 (approved):
LGTM

PR #473, opened 2024-07-30
Use Jazzy image in CI instead of rolling

Ryanf55:
The Jazzy image broke because rolling broke ABI. This should fix it. 

Each time we branch of `rolling`, we need to update the image. `rolling` will only work for a bit.


wep21 (approved):
LGTM

Issue #382, opened 2023-01-13
Eigen Plugins not compiled. 


mariamhegazy:
I want to use grid_map_core so I built my package using colcon but I keep getting this error;
`fatal error: grid_map_core/eigen_plugins/FunctorsPlugin.hpp: No such file or directory`

I am using the repo rolling on ubuntu 22.04 and my cmake file looks like this:

```
cmake_minimum_required(VERSION 3.8)
 project(shaft-mapping)

if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
endif() 
find_package(ament_cmake REQUIRED)
find_package(rclcpp REQUIRED)
find_package(sensor_msgs REQUIRED)
find_package(std_msgs REQUIRED)
find_package(grid_map_core REQUIRED)
find_package(grid_map_cmake_helpers REQUIRED)

if(BUILD_TESTING)
  find_package(ament_lint_auto REQUIRED)
  set(ament_cmake_copyright_FOUND TRUE)
  set(ament_cmake_cpplint_FOUND TRUE)
  ament_lint_auto_find_test_dependencies()
endif()


add_executable(reading_laser src/reading_laser.cpp)
ament_target_dependencies(reading_laser rclcpp std_msgs sensor_msgs)

add_executable(grid src/Grid.cpp)
ament_target_dependencies(grid  grid_map_core)


install(TARGETS
  DESTINATION lib/${PROJECT_NAME}/
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION lib/${PROJECT_NAME}
)

ament_package()
```

samprt-heracles:
Hello, how did you fix this error ?

GradyLiugy:
Hello, how did you fix this error ?

Ryanf55:
Apparently this still affects humble. Re-opened.

StefanFabian:
I've traced it to this cmake file which pollutes the definitions of any project that `find_package`s `grid_map_core`:
https://github.com/ANYbotics/grid_map/blob/master/grid_map_core/cmake/grid_map_core-extras.cmake
here https://github.com/ANYbotics/grid_map/blob/ros2/grid_map_core/CMakeLists.txt#L130

and your `reading_laser` target does not link against `grid_map_core` so it doesn't have the include path for `grid_map_core`.

[...]

PR #475, opened 2024-07-31
Fix Eigen Plugins compilation when linking to grid_map_core

Ryanf55:
# Purpose

Here's a proposed patch for #382. Because the original PR for adding the plugins through CMake and ament is not well documented, I really don't know how this will affect others. Please try this PR out. It will stay open for a while until I get a reasonable number of approvals it doesn't break anything. 

# Test instructions

* Check this branch out
* `colcon build --packages-up-to  issue382`

And try building your own packages that link to grid_map_core.


grid_map_core/CMakeLists.txt
```diff
@@ -5,8 +5,7 @@ project(grid_map_core)
 find_package(ament_cmake REQUIRED)
 find_package(grid_map_cmake_helpers REQUIRED)
 
-## Define Eigen addons.
-include(cmake/${PROJECT_NAME}-extras.cmake)
```
Ryanf55, line 9: This needs to be moved down till after the library is created. You can't call `target_compile_definitions` on a target until the target exists.

grid_map_core/CMakeLists.txt
```diff
@@ -127,6 +129,7 @@ endif()
 
 ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
 ament_export_dependencies(Eigen3)
-ament_package(CONFIG_EXTRAS
-  cmake/${PROJECT_NAME}-extras.cmake
-)
+ament_package()
+#ament_package(CONFIG_EXTRAS
+#  cmake/${PROJECT_NAME}-extras.cmake
```
Ryanf55, line 134: This breaks if we leave it in - any tips on how to test the plugins are working in another package?

nhewitt99:
Just adding my $0.02 that this PR does fix the Eigen linker issues when using grid_map_core for me, thank you! I can't speak to its potential impact on other plugins, though.

Ryanf55:
> Just adding my $0.02 that this PR does fix the Eigen linker issues when using grid_map_core for me, thank you! I can't speak to its potential impact on other plugins, though.

Thanks for sharing!

StefanFabian:
My $0.02: This is in line with what I proposed. I would only move the code from the extas file to the main cmakelists and remove the extras file.
If it is not installed, there's no reason to make it an extra file, and with the naming, it could be confusing.

Generally, setting these Eigen compile time variables in the public interface of a library is IMHO bad style as it limits or at least makes it a lot more complicated for people to add their own extensions and makes it incompatible with any other library that does the same (of which I hope there are none but still).
If it's only used in the cpp files, you could set them as `PRIVATE` and retain the functionality without leaking into downstream packages.

doisyg:
Thanks for this! I am not sure about the proper solution but this is what I have been end up using to be able to build and use properly `grid_map_ros` in external packages (in Jazzy):

https://github.com/botsandus/grid_map/commit/3ce1c30b199718e1d8e5a74014c530c8ed438e15

Ryanf55:
> Thanks for this! I am not sure about the proper solution but this is what I have been end up using to be able to build and use properly `grid_map_ros` in external packages (in Jazzy):
> 
> [botsandus@3ce1c30](https://github.com/botsandus/grid_map/commit/3ce1c30b199718e1d8e5a74014c530c8ed438e15)

THanks for the report. If you take this branch, and try it, does it fix the problem? I think we should merge this, no one said it made things worse.

msplr:
Linking to my proposed fix #519.

cmake extras should not be needed IMO.

PR #476, opened 2024-08-12
Check for already declared parameter

ipa-dmo:
When using the same parameter for several instances of the filter or a filter chain (like in some ros2 forks of your elevation mapping repository), it will throw rclcpp::exceptions::ParameterAlreadyDeclaredException.
This PR just checks if the parameter already exists before declaring it.

### DIFF
--- a/grid_map_cv/include/grid_map_cv/utilities.hpp
+++ b/grid_map_cv/include/grid_map_cv/utilities.hpp
@@ -42,7 +42,10 @@ public:
   {
     rclcpp::Parameter param;
 
-    params_interface_->declare_parameter(param_prefix_ + name, param_type);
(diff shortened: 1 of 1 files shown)


ipa-dmo:
The PR https://github.com/ros/filters/pull/80 in the filterchain base class prevents the rclcpp::exceptions::ParameterAlreadyDeclaredException based on the filter name and type declarations. For the different parameter of the single grid map filters, a check inside your ParameterReader is necessary.

Ryanf55:
This contribution looks good. Any chance you can add a unit test for it to show we no longer throw the `rclcpp::exceptions::ParameterAlreadyDeclaredException` exception?

Ryanf55 (approved):

PR #479, opened 2024-10-17
fix(SpiralIterator): Avoid improper use of '++', which may cause prog…

luanlongshengv:
This PR addresses an issue in the SpiralIterator::operator++() function where the pop_back() method was called on an empty pointsRing_ container, leading to undefined behavior. The fix involves adding a check to ensure that pointsRing_ is not empty before attempting to call pop_back(). This change prevents potential crashes and ensures the iterator operates correctly even when the container is empty. 
This change ensures the iterator’s robustness and prevents potential runtime errors.

### DIFF
--- a/grid_map_core/src/iterators/SpiralIterator.cpp
+++ b/grid_map_core/src/iterators/SpiralIterator.cpp
@@ -49,6 +49,9 @@ const Eigen::Array2i& SpiralIterator::operator *() const
 
 SpiralIterator& SpiralIterator::operator ++()
 {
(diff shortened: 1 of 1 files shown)


Ryanf55 (approved):
Agree, this fixes UB. https://en.cppreference.com/w/cpp/container/vector/pop_back
image

Ryanf55:
@maximilianwulf  Please merge this fix

Ryanf55:
This has been merged into all of the maintained ROS 2 branches.

PR #486, opened 2024-11-26
[ROS 2][grid_map_core] Update ROS 2 branch with the latest changes from master

kjalloul-anybotics:
### Changes to `grid_map_core`

- Added latest fixes and test cases from master.
- Bumped package version to 2.4.0.

### DIFF
--- a/grid_map_core/CHANGELOG.rst
+++ b/grid_map_core/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package grid_map_core
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2.4.0 (2024-11-26)
+------------------
+* Update changes from master
+
 2.3.0 (2024-07-29)
 ------------------
 
@@ -26,6 +30,12 @@ Changelog for package grid_map_core
 ------------------
 * Initial ROS2 port
 * Contributors: Maximilian Wulf, Steve Macenski
+
+1.7.0 (2022-03-17)
+------------------
+
+1.6.4 (2020-12-04)
+------------------
 
 1.6.2 (2019-10-14)
 ------------------
--- a/grid_map_core/include/grid_map_core/GridMap.hpp
+++ b/grid_map_core/include/grid_map_core/GridMap.hpp
@@ -6,8 +6,7 @@
  *	 Institute: ETH Zurich, ANYbotics
  */
 
-#ifndef GRID_MAP_CORE__GRIDMAP_HPP_
-#define GRID_MAP_CORE__GRIDMAP_HPP_
+#pragma once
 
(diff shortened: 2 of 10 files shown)


wep21:
@kjalloul-anybotics rebase, please

Ryanf55:
Thanks for the contribution, nice to see some support for bringing the ROS 2 branch up to date!

 As it stands, you've bundled a number of unrelated changes. Can you instead propose PR's that target things individually so it's more reviewable? Also, bumping the version can be done as it's own PR. 

In addition to the changes outlined, you have
* Replaced ifdefs with `pragma_once`
* Fixed spelling
* Added a union object for the color
* Added test failure messages that don't change the test logic
* Changed CubicInterpolation's underflow behavior
* Cleaned up return value syntax
* Switched to using std::all_of instead of manual iteration
* Switched which exception is thrown in GridMap without updating the public API - this is API breaking and we cannot backport this
... and more...

That's a lot of risk, with no linkage to the original commit history or PR's. Perhaps there's a better strategy we can come up with to make this safer?

kjalloul-anybotics:
Thanks for the feedback!

Yes in the beginning I only needed one new function in `grid_map_cv`, but then while I was at it, the idea was to bring it and its upstream dependency `grid_map_core` up to date with master.

But you're right it definitely can be broken down into smaller PRs to keep track of where the changes came from. 

I traced back the origin of most changes to these commits:
- `std::all_of` changes, return value syntax changes, and other fixes related to formatting from b7293f5d379719d2d0b9f1ce047d00f32601487a
- Typos, updated tests, adding namespace to tests, and more syntax improvements from d6f091225dcc82190a2c24ef12414605e6373b24
- Adding Color union, removing a couple functions, and fixing compiler warnings from 6ddff7413226cdb692055b6686e890d9c034b101
- Clipping point fix and new test cases from f1344bd946967d2806dc9d0e41b119e580398724
- Visibility cleanup from 61972c8732aef46d3cf3a61ad435f9b66a6e65e6
- Transformed map fix + tests from f0dd7800098bdd1fee5630f729b6958c0b091181
- Submap iterator speedup + tests from c8b1057e6d89d485e5fb530e75b00009cd9b1cf6

Do you think it's a good idea to create separate PRs for each of them? Once `grid_map_core` is updated, I can do the same for `grid_map_cv`.

And apparently `pragma`  was replaced with `ifndef` during the ROS 2 port and not the other way around, same with the switch cases to if conditions, so I'll revert these changes.

Ryanf55:
Awesome, thanks for putting the list together!

yes, if it's easy enough, separate PR's are preferred, with a link to the ROS 1 PR.

The change to ifdefs was probably intentional, and no one has complained about it, so I'd personally vote we just not touch it.

PR #492, opened 2024-11-28
Add industrial CI

Ryanf55:
# Purpose

Add the ROS industrial CI which supposedly can help find missing dependencies earlier in a more minimal test environment.

# Details

We currently use `rostooling/setup-ros-docker:ubuntu-noble-ros-rolling-desktop-latest` as an image, which contains all dependencies. 

If we forget dependencies in the package.xml, our CI passes, but the jenkins CI fails.

From what I have read, [ros-industrial](https://github.com/ros-industrial/industrial_ci?tab=readme-ov-file#for-github-actions) can solve that.

Let's see if this can reproduce any test failures we have on the jenkins build farm.

### DIFF
--- a/.github/workflows/industrial-ci.yaml
+++ b/.github/workflows/industrial-ci.yaml
@@ -29,5 +29,11 @@ jobs:
       - name: Set UPSTREAM_WORKSPACE to bring in nav2 as a source dependency
         if: ${{ matrix.env.ROS_DISTRO == 'rolling' }}
         run: echo "UPSTREAM_WORKSPACE=https://raw.githubusercontent.com/ANYbotics/grid_map/rolling/tools/ros2_dependencies.repos" >> $GITHUB_ENV
+      # We duplicate NAV2's key skipping because I doubt they intend this bespoke file to be used by other CI jobs:
+      # https://github.com/ros-navigation/navigation2/blob/main/tools/skip_keys.txt
+      # It also is out of date, so grid_map is just going to skip keys we are sure don't work.
+      - name: set ROSDEP_SKIP_KEYS to skip nav2 keys
+        if: ${{ matrix.env.ROS_DISTRO == 'rolling' }}
+        run: echo "ROSDEP_SKIP_KEYS=slam_toolbox" >> $GITHUB_ENV
       - uses: 'ros-industrial/industrial_ci@master'
         env: ${{matrix.env}}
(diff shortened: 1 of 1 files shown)


Ryanf55:
Ok, the first attempt failed (as I expected) because nav2 is not available as binaries.
https://github.com/ANYbotics/grid_map/actions/runs/12073751899/job/33670483685?pr=492#step:3:242

```
  ERROR: the following packages/stacks could not have their rosdep keys resolved
  to system dependencies:
  grid_map_ros: Cannot locate rosdep definition for [nav2_msgs]
  grid_map_costmap_2d: Cannot locate rosdep definition for [nav2_costmap_2d]
```

Seems like there is a way to bring in other source packages.

https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#workflow

Ryanf55:
Ok, the next commit was able to clone nav2 successfully, however it has errors that slam_toolbox isn't available.

https://github.com/ANYbotics/grid_map/actions/runs/12074002040/job/33671191803?pr=492#step:4:261

The next thing to figure out is how to exclude rosdep keys.

Ryanf55:
If it builds ALL of nav2 (which is unnecessary), we would ideally only build the packages we need in the underlay.

While `colcon build --packages-up-to` would be great, that's not supported. There's good alternatives here:
https://github.com/ros-industrial/industrial_ci/issues/859

For example, in the docs:
```
UPSTREAM_WORKSPACE='https://raw.githubusercontent.com/ros-controls/ros_control/melodic-devel/ros_control.rosinstall -ros_control/rqt_controller_manager'
```

Ryanf55:
Excellent! The addition of the ROS industrial CI was able to isolate the failure we have on the build farm that we didn't see in our CI! 
https://github.com/ANYbotics/grid_map/actions/runs/12074104056/job/33671488356?pr=492#step:5:5021

The workflow is SLOW though. I'd like to merge this as-is, and then do speed optimizations as a follow up.

For speed optimizations, here's some options:
* Only build the packages we need in nav2
* Use mergify to only trigger the workflow when requested as a comment on a PR
* Integrate this as part of a merge pipeline after we click to merge that it runs checks and bounces the PR back if it fails

I don't want this to run every commit, or every time you push to PR because it's so slow.

Ryanf55:
@Mergifyio rebase

Ryanf55:
@mergifyio backport jazzy

Ryanf55:
@Mergifyio backport iron

Ryanf55:
@mergifyio backport humble

PR #493, opened 2024-11-28
Run CI when pushed to Jazzy

Ryanf55:
* We want a build after every merge to the jazzy branch


wep21 (approved):

Ryanf55:
@Mergifyio backport jazzy

PR #494, opened 2024-11-29
[ROS 2][grid_map_core] Reduce clang warnings

kjalloul-anybotics:
Cherry-picked commit b7293f5 (No PR)

Includes fixing typos, changing `typedef` to `using`, `math.h` to `cmath`, `push_back` to `emplace_back`, and other clang warnings.

### DIFF
--- a/grid_map_core/include/grid_map_core/iterators/CircleIterator.hpp
+++ b/grid_map_core/include/grid_map_core/iterators/CircleIterator.hpp
@@ -30,14 +30,14 @@ public:
    * @param center the position of the circle center.
    * @param radius the radius of the circle.
    */
-  CircleIterator(const GridMap & gridMap, const Position & center, const double radius);
+  CircleIterator(const GridMap & gridMap, const Position & center, double radius);
 
   /*!
    * Assignment operator.
    * @param iterator the iterator to copy data from.
    * @return a reference to *this.
    */
-  CircleIterator & operator=(const CircleIterator & other);
+  CircleIterator & operator=(const CircleIterator & other) = default;
 
   /*!
    * Compare to another iterator.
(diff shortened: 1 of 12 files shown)


Ryanf55:
Can you post the clang warnings this solves? Note that the change of the type to a function and the removal of `const` are ABI breaking, so this can only be merged to `rolling`

grid_map_core/include/grid_map_core/iterators/CircleIterator.hpp
```diff
@@ -30,14 +30,14 @@ class CircleIterator
    * @param center the position of the circle center.
    * @param radius the radius of the circle.
    */
-  CircleIterator(const GridMap & gridMap, const Position & center, const double radius);
```
Ryanf55, line 33: I would like know what the motivation is for this change. It goes against const correctness. 
https://isocpp.org/wiki/faq/const-correctness

Can you share what compile warning this solves?

Ryanf55:
One more thing - if this solves all of the clang warnings, it would be fantastic to add a CI job that enforces no new clang warnings. That would be a nice follow up to this.

grid_map_core/include/grid_map_core/iterators/CircleIterator.hpp
kjalloul-anybotics: Ah to be honest I didn't get these warnings myself. The title is a bit confusing, but this was just part of the effort to make rolling as up to date as possible as master.

Is it not advisable to fix these if we're not seeing them ourselves during `colcon build`? Then I would just port the major changes from [these](https://github.com/ANYbotics/grid_map/pull/486#issuecomment-2505618400), but it's a bit confusing to know what was a needed fix and what was just a code clean up.

Ryanf55: Yea, especially if the change breaks ABI, I don't see any strong reason to fold it in. Perhaps it's worth starting with any bugfixes that have associated tests?

PR #503, opened 2024-12-03
[ROS 2][grid_map_core] Code enhancements from master 6ddff74

kjalloul-anybotics:
Cherry-picked commit 6ddff74 (No PR)

Includes:
- Adding `Color` union
- Removing unused function `getIndicesForRegion`
- Using `static_cast` instead of c-style cast
- Fixing unused function parameters


kjalloul-anybotics:
@Ryanf55 Not sure if removing `getIndicesForRegion` counts as a breaking change here.

The functions were already commented out in `GridMapMath.cpp` and unused in the repository.

PR #507, opened 2024-12-03
[ROS 2][grid_map_core] Speed up submap iterator c8b1057

kjalloul-anybotics:
Cherry-picked commit c8b1057 (No PR)

Speeds up the submap iterator and adds new test cases.


kjalloul-anybotics:
> Nice! Did the original work have benchmark proof that could be attached?

It doesn't look like it. All I could find was the commit c8b1057 but no PR or issue for it.

PR #515, opened 2025-02-21
Fix the README.md (Foxy --> Humble)

teyssieuman:
The `README.md` was still reffering to the ROS2 Foxy  distribution in this Humble branch.


Ryanf55:
Can you change this to target rolling, and then say it's been tested on Humble and Jazzy?

PR #519, opened 2025-04-30
[ROS2] Restrict Eigen plugin definitions to only targets linking against grid_map_core

msplr:
This addresses #382 such that the Eigen plugin include paths are only defined for targets linking against grid_map_core.

I manually tested this in a dummy package with two executables, where only one is linking against grid_map_core.

@Ryanf55 Should this fix target rolling and/or jazzy?


grid_map_core/CMakeLists.txt
```diff
@@ -127,6 +130,4 @@ endif()
 
 ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
 ament_export_dependencies(Eigen3)
-ament_package(CONFIG_EXTRAS
-  cmake/${PROJECT_NAME}-extras.cmake
```
msplr, line 131: The cmake extras should not be needed, since the public `target_compile_definitions` will be propagated to all dependencies.

Ryanf55:
Thanks, looks good. Rolling is perfect.
I'll need to troubleshoot CI before we can merge this. If you have suggestions for getting that back up, it would be helpful because I'm slammed at work the next two weeks.

Ryanf55:
This seemed to introduce problems in the build. Help is needed. I don't have enough time to chase the new failures down.
https://github.com/Ryanf55/grid_map/pull/5

PR #520, opened 2025-05-07
grid_map: fix compile issues with ROS2

sssarana:
- include correct header files for `cv_bridge`;
- updated CMake files to properly add dependencies;
- updated CMake to add definitions for `eigen` plugins correctly;
- add dependency on `grid_map_core` for demos and rviz plugin;
- corrected outdated use of `send_timestamp` in ROS converter;


grid_map_cv/CMakeLists.txt
```diff
@@ -43,13 +43,8 @@ add_library(${PROJECT_NAME} SHARED
...
 
-target_link_libraries(${PROJECT_NAME} PUBLIC
-  cv_bridge::cv_bridge
-  grid_map_core::grid_map_core
-  opencv_photo
-  pluginlib::pluginlib
-  rclcpp::rclcpp
-  ${sensor_msgs_TARGETS}
+ament_target_dependencies(${PROJECT_NAME} SYSTEM
+  ${dependencies}
```
Ryanf55, line 47: This is backwards, we should link to exported namespace targets. If any are missing, add them.

grid_map_ros/test/GridMapRosTest.cpp
```diff
@@ -14,7 +14,7 @@
 #include <Eigen/Core>
 
 // ROS
-#include <cv_bridge/cv_bridge.hpp>
+#include <cv_bridge/cv_bridge.h>
```
Ryanf55, line 17: I think your library is out of date. rolling uses the hpp extension.
https://github.com/ros-perception/vision_opencv/blob/rolling/cv_bridge/include/cv_bridge/cv_bridge.hpp

Ryanf55 (changes requested):
Sorry, I don't think all the patches here are right.
Please submit separate patches for the things that are still valid.

PR #529, opened 2025-12-12
grid_map_sdf: fix planar SDF memory leak

Idate96:
leak in SignedDistanceField::getPlanarSignedDistanceField(): the temporary distance_transform::image<uchar> and the dt() output were allocated with new and never freed. This switches to RAII (stack allocation + std::unique_ptr) so repeated SDF recomputation doesn't grow memory.


Ryanf55:
HI, can you please retarget this to rolling? LGTM.

PR #532, opened 2026-03-14
feat(grid_map_visualization): convert GridMapVisualization to rclcpp_lifecycle::LifecycleNode

Lakshya-04:
## Motivation

`GridMapVisualization` runs continuously even when no operator has RViz open, paying the full cost of deserializing every incoming `GridMap` message and converting it to point clouds / marker arrays for nobody. The existing implementation works around this with a 2 Hz wall timer that polls subscriber counts and lazily subscribes/unsubscribes — but this is inherently racy and adds its own timer overhead.

A lifecycle node solves this cleanly: an external lifecycle manager (e.g. `nav2_lifecycle_manager`, a mission controller, or a simple launch-time `ros2 lifecycle set` call) activates the visualization only when a human is monitoring and deactivates it when not needed.

## Changes

**`GridMapVisualization`** — inherits from `rclcpp_lifecycle::LifecycleNode` instead of owning an internal `rclcpp::Node`:
- `on_configure`: reads parameters, constructs `VisualizationFactory` and all configured visualizations
- `on_activate`: creates the `GridMap` subscription and starts processing incoming maps
- `on_deactivate`: resets the subscription — zero CPU spent on visualization when inactive
- `on_cleanup` / `on_shutdown`: tears down visualizations and factory

Removes: `activityCheckTimer_`, `isSubscribed_`, `updateSubscriptionCallback()`, and the public `nodePtr` member.

`VisualizationBase`, `VisualizationFactory`, and all concrete visualization classes are **unchanged** — they accept `rclcpp::Node::SharedPtr` and receive one via `std::static_pointer_cast` from the lifecycle node.

**`grid_map_visualization_node.cpp`** — spins via `get_node_base_interface()` (standard lifecycle node spin pattern).

**`package.xml` / `CMakeLists.txt`** — add `rclcpp_lifecycle` dependency.

## Behaviour change

The node now starts in the `unconfigured` state. It must be transitioned to `active` before it processes any messages:

```bash
ros2 lifecycle set /grid_map_visualization configure
ros2 lifecycle set /grid_map_visualization activate
```

Or managed automatically via `nav2_lifecycle_manager` in a launch file.

## Test plan

- Verified node transitions correctly through `configure → active → inactive → cleanup` via `ros2 lifecycle set`
- Confirmed point cloud and occupancy grid visualizations publish when active
- Confirmed no messages are processed and no subscription exists when inactive
- Tested on ROS 2 Humble

### DIFF
--- a/grid_map_visualization/CMakeLists.txt
+++ b/grid_map_visualization/CMakeLists.txt
@@ -9,6 +9,7 @@ find_package(grid_map_msgs REQUIRED)
 find_package(grid_map_ros REQUIRED)
 find_package(nav_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
 find_package(sensor_msgs REQUIRED)
 find_package(visualization_msgs REQUIRED)
 find_package(Eigen3 REQUIRED)
@@ -21,6 +22,7 @@ set(dependencies
   grid_map_ros
   nav_msgs
   rclcpp
+  rclcpp_lifecycle
   sensor_msgs
   visualization_msgs
 )
--- a/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
+++ b/grid_map_visualization/include/grid_map_visualization/GridMapVisualization.hpp
@@ -19,6 +19,7 @@
 
 // ROS
 #include <rclcpp/rclcpp.hpp>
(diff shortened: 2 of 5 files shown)


Ryanf55:
Overall, I like lifecycle nodes, but I don't see how we can merge this to humble because it breaks public ABI.
Is there any way you'd consider changing this to rolling? We need to deprecate calls for a release cycle before we can just delete them.

PR #542, opened 2026-07-19
Match grid_map_pcl timer clock type

Tobias-Fischer:
This is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: patch/ros-rolling-grid-map-pcl.osx.patch, authored by Daisuke Nishimatsu.

Best-guess rationale: processPointcloud records the start time with std::chrono::high_resolution_clock::now(), but the helper accepted a system_clock time_point. Matching the helper parameter to the clock used by the caller avoids implementation-dependent clock type mismatches, including libc++ configurations where high_resolution_clock is not system_clock.


Tobias-Fischer:
FYI @wep21 and @traversaro: this upstreams a RoboStack downstream patch tracked in https://github.com/RoboStack/robostack.github.io/issues/16.

